PR #5, opened 2021-04-26
add filter for good pairs and function for the four vectors

harrypuuter:


pairselection.hxx
```diff
@@ -16,6 +16,14 @@
 
 namespace pairselection {
 
+    auto filterGoodParticles(auto df, const std::string& pairname, const std::string& filtername){
```
ArturAkh, line 19: I would suggest to rename it to `filterGoodPair`, since it could be misunderstood.

Another point for my understanding: For which purpose is `const std::string& filtername` needed? To name the filter for the report, right?

lorentzvectors.hxx
```diff
@@ -0,0 +1,45 @@
...
+                // auto p4s = ROOT::VecOps::Construct<ROOT::Math::PtEtaPhiMVector>(pts, etas, phis, masses);
+                //return p4s[index];
+                Logger::get("lorentzvectors")->debug("pair {}", pair);
+                Logger::get("lorentzvectors")->debug("pts {}", pts);
+                Logger::get("lorentzvectors")->debug("etas {}", etas);
+                Logger::get("lorentzvectors")->debug("phis {}", phis);
+                Logger::get("lorentzvectors")->debug("masses {}", masses);
+                Logger::get("lorentzvectors")->debug("Index {}", index);
+
+                const auto p4 = ROOT::Math::PtEtaPhiMVector(pts[index], etas[index], phis[index], masses[index]);
```
ArturAkh, line 27: I guess we don't use `RVec` of `PtEtaPhiMVector` because it is always one particle?

analysis.cxx
```diff
@@ -56,18 +56,24 @@ int main(){
...
 
     varSet.push_back("good_muons_mask");
     varSet.push_back("good_taus_mask");
     varSet.push_back("mtpair");
+    varSet.push_back("particle_1_p4");
+    varSet.push_back("particle_2_p4");
     Logger::get("main")->info("Finished Setup");
     Logger::get("main")->info("Starting Evaluation");
 
-    df_final.Snapshot<ROOT::VecOps::RVec<int>, ROOT::VecOps::RVec<int>, ROOT::VecOps::RVec<int>>("ntuple", "test.root", varSet);
```
ArturAkh, line 70: That's fine with me for the time-being.

However, as soon as we have a full list of quantities for the `mutau` pair to be written out, I'd suggest to test the non-jitted vs. jitted case.

analysis.cxx
```diff
@@ -56,18 +56,24 @@ int main(){
     // Build the pair !
     auto df24_1 = physicsobject::FilterObjects(df24, "nTau", 1,"NumberOfTaus");
     auto df24_2 = physicsobject::FilterObjects(df24_1, "nMuon", 1,"NumberOfMuons");
-    auto df25 = pairselection::mutau::PairSelection(df24_2, "good_taus_mask", "good_muons_mask", "mtpair", {""});
-    auto df_final = df25;
+    auto mt_df = pairselection::mutau::PairSelection(df24_2, "good_taus_mask", "good_muons_mask", "mtpair", {""});
+    auto mt_df_1 = pairselection::filterGoodParticles(mt_df, "mtpair", "GoodMuTauPairs");
```
ArturAkh, line 60: Suggest to rename to `filterGoodPair` for better understanding (please see also comment below in the `pairselection.cxx`)

ArturAkh (changes requested):
Only a minor change of a filter function name & some questions for my understanding added.

Everything else looks good :)

pairselection.hxx
harrypuuter: yes its for the report - i will rename it

lorentzvectors.hxx
harrypuuter: yes, we only have to build one particle anyways, and i would also keep this split for the tautau channel, so re can reuse the code

ArturAkh (approved):

PR #6, opened 2021-04-26
add infrastructure in cmake to generate files

stwunsch:


ArturAkh (approved):
Since you have tested that, @stwunsch, I would suggest to merge that directly into the main branch of the repo :)

We can adapt it then afterwards on the main branch

PR #16, opened 2021-04-30
Refactored the docs branch

stwunsch:


stwunsch:
@harrypuuter The full docs are now in `${CMAKE_BINARY_DIR}/docs` and can just be copied as it is to a webserver. Just have to figure out which webserver ;)

stwunsch:
I would not publish it statically and commit it to this repo, this would clutter it massively on the long run. Perhaps we can make another repo as a plain gh-pages target?

rcaspart:
Without having had a closer look at the docs and code, did you consider readthedocs?

stwunsch:
> Without having had a closer look at the docs and code, did you consider readthedocs?

Good idea! It looks like we can even build it there? Well, having the checks as part of the CI is also not bad. But indeed, readthedocs could be at least used to host this!

stwunsch:
I have never seen C++ documentation on readthedocs 🤔

rcaspart:
HTCondor is on readthedocs. Not sure how they build their docs though.

harrypuuter:
the reason why I build the system to end up with the final documentation in /docs is because github supports github pages from the main directory or the /docs folder. But Indeed, using redthedocs would be maybe a nicer alternative, since then we are not dependent on the location of where the docs are stored and this would not pollute the repo

harrypuuter:
I added the stuff needed for readthedocs

stwunsch:
There still (or again) quite a number of dotfiles hanging around in `docs/`!

harrypuuter (approved):

PR #17, opened 2021-04-30
python backbone

swozniewski:
Develops python backbone such that current example analysis.cxx can be reproduced.
In addition, an exemplary shift is introduced. Note, that filtering with shift needs better treatment.
Channel specific treatment not yet implemented.
Some cpp modules are modified to better fit a common structure.

### DIFF
--- a/code_generation/producer.py
+++ b/code_generation/producer.py
@@ -260,6 +260,6 @@ UnrollLV1 = ProducerGroup(None, None, None, [pt_1, eta_1, phi_1])
 UnrollLV2 = ProducerGroup(None, None, None, [pt_2, eta_2, phi_2])
 
 m_vis = Producer(
-    'quantities::pt({df}, varSet, "{output}", {input_coll})', [q.p4_1, q.p4_2], q.m_vis
+    'quantities::m_vis({df}, varSet, "{output}", {input_coll})', [q.p4_1, q.p4_2], q.m_vis
 )
 DiTauPairQuantities = ProducerGroup(None, None, None, [UnrollLV1, UnrollLV2, m_vis])
--- a/code_generation/producer.py
+++ b/code_generation/producer.py
@@ -17,13 +17,17 @@ class Producer:
             q.children.append(self.output)
 
     def shift(self, name):
-        self.output.shift(
-            name
-        )  # crashes on purpose if output is None. This method should not be called for a producer without output
+        if isinstance(self.output, list):
+            for entry in self.output:
+                entry.shift(name)
+        else:
+            self.output.shift(
+                name
+            )  # crashes on purpose if output is None. This method should not be called for a producer without output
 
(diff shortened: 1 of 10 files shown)


quantities.hxx
```diff
@@ -2,41 +2,42 @@
...
 
 namespace quantities {
-auto pt(auto df, std::vector<std::string> varSet, const std::string &outputname,
-        const std::string &inputvector) {
+auto pt(auto df, std::vector<std::string> &varSet,
+        const std::string &outputname, const std::string &inputvector) {
     varSet.push_back(outputname);
     return df.Define(
         outputname,
         [](const ROOT::Math::PtEtaPhiMVector &p4) { return p4.pt(); },
```
stwunsch, line 10: Hier muesst ihr aufpassen, das schreibt euch ein double raus und ihr habt am Ende evtl ein 2x so grosses ntuple ;) Selbes gilt auch fuer die anderen quantities unten.

```suggestion
        [](const ROOT::Math::PtEtaPhiMVector &p4) { return (return) p4.pt(); },
```

code_generation/producer.py
```diff
@@ -13,17 +13,26 @@ def __init__(self, call, inputs, output):
         self.output = output
 
         # keep track of variable dependencies
-        for q in self.inputs:
-            q.children.append(self.output)
+        if self.output != None:
+            for qy in self.inputs:
```
harrypuuter, line 17: `qy` should be `quantity` ?

code_generation/producer.py
```diff
@@ -37,47 +46,222 @@ def writecall(self, config, shift=""):
...
         calls = []
         shifts = [""]
         if self.output != None:
-            shifts.extend(self.output.shifts)
+            shifts.extend(self.output[0].shifts)
         for shift in shifts:
-            for val in config[shift][self.vec_config]:
-                self.call = basecall.format_map(SafeDict({self.vec_config: val}))
+            # check that all config lists (and output if applicable) have same length
+            l = len(config[shift][self.vec_configs[0]])
```
harrypuuter, line 72: switch from `l` to some longer more descriptive variable name

code_generation/producer.py
```diff
@@ -37,47 +46,222 @@ def writecall(self, config, shift=""):
...
+    PG_count = 1  # counter for internal quantities used by ProducerGroups
+
+    def __init__(self, call, inputs, output, subproducers):
+        self.call = call
+        self.inputs = inputs
+        self.output = output
+        self.producers = subproducers
+        # If call is provided, this is supposed to consume output of subproducers. Creating these internal products below:
+        if self.call != None:
+            for p in self.producers:
```
harrypuuter, line 107: `for producer in self.producers`

harrypuuter (changes requested):

harrypuuter (approved):

PR #20, opened 2021-05-04
add producer names and include producers via config

swozniewski:


code_generation/code_generation.py
```diff
@@ -1,29 +1,42 @@
...
     commandlist = ""  # string to be placed into code template
     df_count = 0  # enumerate dataframes
+    df_scope_count = 0  # enumerate dataframes in certain scopes
 
     # get commands of producers and append to the command list
-    for producer in config["producers"]:
-        commandlist += "\n    //" + producer + "\n"
-        for call in getattr(p, producer).writecalls(config):
-            # print(call)
+    for producer in config["producers"]["global"]:
```
harrypuuter, line 13: Couldnt this function be put in the for loop below, there we iterate on global anyawys

code_generation/producer.py
```diff
@@ -142,126 +161,200 @@ def writecalls(self, config):
...
-    'quantities::phi({df}, varSet, "{output}", "{input}")', [q.p4_2], q.phi_2
+    "phi_2",
+    'quantities::phi({df}, varSet, "{output}", "{input}")',
+    [q.p4_2],
+    q.phi_2,
+    ["mt"],
 )
-UnrollLV1 = ProducerGroup(None, None, None, [pt_1, eta_1, phi_1])
-UnrollLV2 = ProducerGroup(None, None, None, [pt_2, eta_2, phi_2])
+UnrollLV1 = ProducerGroup("UnrollLV1", None, None, None, ["mt"], [pt_1, eta_1, phi_1])
```
harrypuuter, line 348: None, None None looks a bit stange here, either we add the keywords, or maybe change the ordering of the paramteters. From just looking at this line, its not clear at all what these Nones do

code_generation/quantities.py
```diff
@@ -1,27 +1,52 @@
...
+                    for c in self.children[scope]:
+                        c.shift(name, scope)
 
     def copy(self, name):
         copy = Quantity(name)
         copy.shifts = self.shifts
         copy.children = self.children
         return copy
 
+    def adopt(self, child, scope):
```
harrypuuter, line 33: adopt --> adopt_child

harrypuuter (changes requested):

code_generation/code_generation.py
swozniewski: It must be done before all other scopes and I think the dict sorts alphabetically and would change the order given in the config.

code_generation/producer.py
swozniewski: Right, you suggested to add the keywords once before. That's probably a good thing and I will add it.

code_generation/quantities.py
swozniewski: What else should it adopt? I'd prefer to keep it short. It's anyway not to be used by main stream users.

harrypuuter: yes it was a translation error on my side (adapt <--> adopt) :)

harrypuuter (approved):
nice !

PR #23, opened 2021-05-05
Configure snapshots via python part

swozniewski:
Content of snapshots is defined in config per scope.
Snapshots are created for each scope that the content is configured for.
All shifts of respective quantities within the given scope are automatically added to the content.

### DIFF
--- a/config/config.py
+++ b/config/config.py
@@ -1,4 +1,5 @@
 import code_generation.producer as p
+from code_generation.producer import q
 from config.utility import AddSystematicShift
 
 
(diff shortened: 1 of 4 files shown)


config/config.py
```diff
@@ -40,6 +40,13 @@ def build_config():
         ],
     }
 
+    config["output"] = {
+        "mt": [
+            p.q.pt_1,
```
harrypuuter, line 45: this looks strange, either:
```
producer.q.pt_1
```
or 
```
p.quantity.pt_1
```
or even 
```
producer.quantity.pt_1
```

harrypuuter (changes requested):

config/config.py
swozniewski: Maybe import p.q separately as q?

swozniewski: done. Please check again.

harrypuuter (approved):

PR #47, opened 2021-05-18
Jets

swozniewski:


jets.hxx
```diff
@@ -0,0 +1,105 @@
+#include "ROOT/RDataFrame.hxx"
+#include "ROOT/RVec.hxx"
+#include "basefunctions.hxx"
+#include "utility/Logger.hxx"
+#include <Math/Vector3D.h>
+#include <Math/Vector4D.h>
+#include <Math/VectorUtil.h>
+
+namespace jet {
+auto VetoOverlappingJets(auto df, const std::string &output_col,
```
harrypuuter, line 10: We need documentation here

jets.hxx
```diff
@@ -0,0 +1,105 @@
...
+                    ->debug("DeltaR 2 {}", deltaR_2);
+                mask[idx] = (deltaR_1 > deltaRmin && deltaR_2 > deltaRmin);
+            }
+            return mask;
+        },
+        {jet_eta, jet_phi, p4_1, p4_2});
+    return df1;
+}
+
+auto OrderJetsByPt(auto df, const std::string &output_col,
```
harrypuuter, line 46: Documentation

jets.hxx
```diff
@@ -0,0 +1,105 @@
...
+            auto good_jets_pt =
+                ROOT::VecOps::Where(jetmask > 0, jet_pt, (float)0.);
+            Logger::get("OrderJetsByPt")->debug("Jetpt after {}", good_jets_pt);
+            // we have to convert the result into an RVec of ints since argsort
+            // gives back an unsigned long vector
+            auto temp =
+                ROOT::VecOps::Argsort(ROOT::VecOps::Nonzero(good_jets_pt));
+            Logger::get("OrderJetsByPt")->debug("jet Indices {}", temp);
+            ROOT::RVec<int> result(temp.size());
+            std::transform(temp.begin(), temp.end(), result.begin(),
```
harrypuuter, line 63: This part did cause some crashes, when i first did this, but i guess this is fixed now ?

jets.hxx
```diff
@@ -0,0 +1,105 @@
...
+              const int idxID) {
+    auto df1 = df.Define(maskname, basefunctions::FilterJetID(idxID), {nameID});
+    return df1;
+}
+} // end namespace jet
+} // end namespace physicsobject
+
+namespace quantities {
+namespace jet {
+auto NumberOfJets(auto df, const std::string &outputname,
```
harrypuuter, line 94: Documentation

quantities.hxx
```diff
@@ -47,10 +48,13 @@ auto eta(auto df, const std::string &outputname,
 
 auto phi(auto df, const std::string &outputname,
          const std::string &inputvector) {
-    return df.Define(
-        outputname,
-        [](const ROOT::Math::PtEtaPhiMVector &p4) { return (float)p4.phi(); },
-        {inputvector});
+    return df.Define(outputname,
+                     [](const ROOT::Math::PtEtaPhiMVector &p4) {
+                         if (p4.pt() < 0.0)
```
harrypuuter, line 53: Is pt negative for the default lorentz vector ? --> add a comment

harrypuuter: also is there any speed difference between using if..else and the conditional operator as we did it in artus ? - if its the same, then i guess we keep if else

harrypuuter (changes requested):

jets.hxx
swozniewski: Right, this is what I mentioned in my private message already. The transform function does not check the size of the target vector (it was initialized empty before). Now it is initialized with correct size. There is also a method to increase size within transform (inserting some function similar to push_back instead of the plain iterator), but I think it's best to reserve the correct amount of memory right away.

quantities.hxx
swozniewski: Yes, I checked that pt is still the same vector as in the initialization. Therefore, I think it is ok to use negative pt values as markers that four-vectors are not properly filled and default values should be used. (Alternatively the length of the corresponding mask or njets could always be checked but that requires to pipe an additional quantity through the functions...)
While defaults can be inserted for eta and pt, the phi value is taken modulo 2pi and therefore the default value must be set manually for retrieving phi.
I don't know whether there are speed differences. I would be surprised if they were measaurably large but we can try to find out...

harrypuuter (approved):

PR #48, opened 2021-05-26
add new Quantities

harrypuuter:
This PR adds writeout for 

- mass
- dxy
- dx
- q 
- iso
- decaymode
- genmatch


code_generation/producers/ditaupairquantities.py
```diff
@@ -0,0 +1,179 @@
...
+    name="m_vis",
+    call="quantities::m_vis({df}, {output}, {input_vec})",
+    input=[q.p4_1, q.p4_2],
+    output=[q.m_vis],
+    scopes=["mt", "et", "tt", "em"],
+)
+####################
+# Set of channel specific producers
+####################
+mass_1 = Producer(
```
swozniewski, line 60: this naming will lead to collisions when other channels are intoduced with a corresponding quantity. Can't we take it from the 4-Vector, so it could be the same definition for all channels?

code_generation/producers/ditaupairquantities.py
```diff
@@ -0,0 +1,179 @@
...
+    name="mass_2",
+    call="quantities::mass({df}, {output}, 1, {input})",
+    input=[q.ditaupair, q.Tau_mass],
+    output=[q.mass_2],
+    scopes=["mt", "et", "tt"],
+)
+dxy_1 = Producer(
+    name="dxy_1",
+    call="quantities::dxy({df}, {output}, 0, {input})",
+    input=[q.ditaupair, q.Muon_dxy],
```
swozniewski, line 77: I'd again suggest to hardcode such nanoAOD quantities in the call (unless certain difficulties arise, like a case we had with jets I think). Similar for most quantities below.

quantities.hxx
```diff
@@ -57,8 +57,92 @@ auto phi(auto df, const std::string &outputname,
...
+/// \param df the dataframe to add the quantity to
+/// \param outputname name of the new column containing the mass value
+/// \param position index of the position in the pair vector
+/// \param pairname name of the column containing the pair vector
+/// \param masscolum name of the column containing the mass values
+///
+/// \returns a dataframe with the new column
+
+auto mass(auto df, const std::string &outputname, const int position,
+          const std::string &pairname, const std::string &masscolum) {
```
swozniewski, line 72: colum->column here and everywhere else in the file

swozniewski (changes requested):

code_generation/producers/ditaupairquantities.py
harrypuuter: yes, i will do this in a seperate PR for all existing producers at once

quantities.hxx
```diff
@@ -57,8 +57,92 @@ auto phi(auto df, const std::string &outputname,
                      },
                      {inputvector});
 }
-/// Function to calculate the mass from a pair of lorentz vectors and add it to
-/// the dataframe
+/// Function to writeout the mass from a particle. The particle is identified
```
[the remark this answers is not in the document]
harrypuuter: The thing is, that the function has to be different depending on the datatype of the output, e.g. charge is an int, genmatch is a uint etc. This would mean, that we need several of those generic functions, and as soon as something more complicated has to be done, these basic functions wont work anymore. So i would suggest to keep it this way, with one function per quantity, even if this results in some duplications, for consistency reasons.

swozniewski (approved):

PR #49, opened 2021-05-27
add check for output quantity usage

harrypuuter:
This PR adds 

* a check for the double usage of output quantities
* a consistent separation between NanoAOD quantities and new quantities
* restructuring of files, such that the classes and their instances are separated


code_generation/code_generation.py
```diff
@@ -13,10 +13,10 @@ def fill_template(t, config):
     commandlist = ""  # string to be placed into code template
     df_count = 0  # enumerate dataframes
     df_scope_count = 0  # enumerate dataframes in certain scopes
-
     # get commands of producers and append to the command list
     log.info("Generating commands ...")
     for producer in config["producers"]["global"]:
+        producer.check_output()
```
swozniewski, line 19: I'd prefer a name like 'reserve_output' to make clear that it changes something and does not only do a look up.

code_generation/producer.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, name, call, input, output, scopes):
         self.input = input
         self.output = output
         self.scopes = scopes
+        self.output_checked = False
```
swozniewski, line 28: I think this attribute is not needed and can be removed. Such tracking is only needed for quantities.

code_generation/producer.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, name, call, input, output, scopes):
         log.debug("| scopes: {}".format(self.scopes))
         log.debug("-----------------------------------------")
 
+    def check_output(self):
+        if self.output != None and not self.output_checked:
```
swozniewski, line 50: remove second condition

code_generation/code_generation.py
swozniewski: the scope will be necessary as argument

code_generation/producer.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, name, call, input, output, scopes):
...
+            for scope in self.scopes:
```
swozniewski, line 51: do not loop over scopes here. Take the scope by argument and only apply for that scope. The producer might be used multiple times in different pipelines such that this function will be called a second time and crash. Also, if the producer is not run in all scopes it is designed for, you will erroneously mark the quantity as available in that scope.

code_generation/producer.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, name, call, input, output, scopes):
...
+                for output_quantity in self.output:
+                    output_quantity.check_scope(scope)
+        self.output_checked = True
```
swozniewski, line 54: remove

code_generation/producer.py
```diff
@@ -45,6 +46,13 @@ def __init__(self, name, call, input, output, scopes):
         log.debug("| scopes: {}".format(self.scopes))
         log.debug("-----------------------------------------")
 
+    def check_output(self):
+        if self.output != None and not self.output_checked:
+            for scope in self.scopes:
+                for output_quantity in self.output:
+                    output_quantity.check_scope(scope)
```
swozniewski, line 53: better call the method 'reserve'. That it is scope specific is clear by the argument.

code_generation/producer.py
```diff
@@ -215,6 +223,10 @@ def __init__(self, name, call, input, output, scopes, subproducers):
         log.debug("| scopes: {}".format(self.scopes))
         log.debug("-----------------------------------------")
 
+    def check_output(self):
+        for subproducer in self.producers:
+            subproducer.check_output()
```
swozniewski, line 228: adapt naming as above

code_generation/quantity.py
```diff
@@ -0,0 +1,91 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class Quantity:
+    def __init__(self, name):
+        self.name = name
+        self.shifts = {}
+        self.children = {}
+        self.output_scopes = []
```
swozniewski, line 11: I'd call it self.defined_for_scopes. 'output' does not make sense to me.

code_generation/quantity.py
```diff
@@ -0,0 +1,91 @@
...
+        log.debug("Setting up new Quantity {}".format(self.name))
+
+    # the scopes, in which a quantity is used as an output is tracked in the output_scopes list.
+    # This check is triggered for every producer.
+    # If a quantity is already used within a given scope as output, this will result in an exception.
+    def check_scope(self, scope):
+        log.debug("Checking {} / scope {}".format(self.name, scope))
+        if scope not in self.output_scopes:
```
swozniewski, line 19: you should also check whether it global is present

code_generation/quantity.py
```diff
@@ -0,0 +1,91 @@
...
+            self.output_scopes.append(scope)
+        else:
+            log.error(
+                "Quantity {} is already used as output in {} scope !".format(
```
swozniewski, line 23: used as output -> defined (sounds clearer to me)

code_generation/quantity.py
```diff
@@ -0,0 +1,91 @@
...
+            if "global" in self.shifts.keys():
+                return list(self.shifts["global"])
+            else:
+                return []
+
+
+class NanoAODQuantity(Quantity):
+    def __init__(self, name):
+        self.name = name
+        super().__init__(name)
```
swozniewski, line 82: Is this needed?

code_generation/producer.py
harrypuuter: yes, i had this initially due to the iterative nature of ProducerGroups, but adding the reserve function for the Group as well does make this obsolete

code_generation/quantity.py
harrypuuter: renamed

code_generation/quantity.py
harrypuuter: but global is also just a scope or am i misunderstanding your comment ?

code_generation/quantity.py
harrypuuter: changed

code_generation/quantity.py
harrypuuter: nope

swozniewski (approved):

PR #52, opened 2021-05-31
Tau es

swozniewski:


code_generation/producer.py
```diff
@@ -68,9 +68,23 @@ def shift(self, name, scope="global"):
...
             raise Exception
         for entry in self.output:
-            entry.shift(
-                name, scope
-            )  # crashes on purpose if output is None. This method should not be called for a producer without output
+            entry.shift(name, scope)
+
+    def ignore_shift(self, name, scope="global"):
+        if not scope in self.scopes:
+            print(
```
harrypuuter, line 75: use logging instead of print

code_generation/producer.py
```diff
@@ -68,9 +68,23 @@ def shift(self, name, scope="global"):
...
+                "Trying to add shift %s to producer %s in scope %s, but producer does not exist in given scope!"
+                % (name, self.name, scope)
+            )
+            raise Exception
+        if self.output is None:
+            print(
```
harrypuuter, line 81: logging

code_generation/quantity.py
```diff
@@ -33,20 +34,30 @@ def get_leafs_of_scope(self, scope):
         return [self.name] + [self.name + shift for shift in self.get_shifts(scope)]
 
     def shift(self, name, scope):
+        if scope in self.ignored_shifts.keys():
+            if name in self.ignored_shifts[scope]:
+                log.debug("Ignoring shift {} for quantity {}".format(name, self.name))
+                return
```
harrypuuter, line 40: This is a bit inconsistent, in this case the function returns a None, and in all other cases nothing is returned. Either all possible cases should return something, or no case should return something. So i would suggest to remove the return and use an else for the other cases

harrypuuter (changes requested):

code_generation/quantity.py
swozniewski: I think a python function without return statement returns None as well.

harrypuuter (approved):

PR #56, opened 2021-06-10
Filter

swozniewski:


code_generation/producer.py
```diff
@@ -271,3 +273,13 @@ def writecalls(self, config, scope):
             # retrieve calls of subproducers
             calls.extend(producer.writecalls(config, scope))
         return calls
+
+
+class Filter(ProducerGroup):
```
harrypuuter, line 278: Should this be a Group Filter, rather than a filter only ?

basefunctions.hxx
```diff
@@ -14,6 +14,29 @@ enum Channel { MT = 0, ET = 1, TT = 2, EM = 3 };
 
 namespace basefunctions {
 
+/// This function filters events, where neither of the input flags is true.
+/// This is used to filter events which do not pass an underlying requirement in
+/// any systematic variation.
+///
+/// \param df The input dataframe
+/// \param filtername The name of the filter, used in the Dataframe report
+/// \param flags Parameter pack of column names that contain the considered
```
harrypuuter, line 23: add a sentence, that these filter flags are requrired to be bools

pairselection.hxx
```diff
@@ -48,22 +48,24 @@ auto buildgenpair(auto df, const std::string &recopair,
...
+/// \param flagname The name of the generated flag column
 /// \param pairname The name of the column, containing the particle indices
-/// \param filtername The name of the filter, used in the Dataframe report
 /// index of the particle in the particle quantity vectors.
 ///
-/// \returns a filtered dataframe
-auto filterGoodPairs(auto df, const std::string &pairname,
-                     const std::string &filtername) {
+/// \returns a dataframe with the new flag
+auto filterGoodPairs(auto df, const std::string &flagname,
```
harrypuuter, line 62: maybe we change the naming to something like `flagGoodPairs`, since we do not filter here, we just flag the events

basefunctions.hxx
```diff
@@ -14,6 +14,29 @@ enum Channel { MT = 0, ET = 1, TT = 2, EM = 3 };
...
+/// flags
+///
+/// \returns a filtered dataframe
+template <class... Flags>
+auto FilterPassAny(auto df, const std::string filtername,
```
harrypuuter, line 28: Maybe change name to `FilterFlags` - in conjunction with `flagGoodPairs`

harrypuuter (changes requested):

code_generation/producer.py
swozniewski: yes, I thoght that would be useful since every filter will contain the actual DF-Filterstep and a producer of the input flags. This saves us an instance of ProducerGroup that just collects the two parts.

pairselection.hxx
swozniewski: agreed

basefunctions.hxx
swozniewski: maybe FilterFlagsAny or FilterFlagsOR?

harrypuuter: i like `FilterFlagsAny` :+1:

code_generation/producer.py
harrypuuter: Should we also change this to `FilterFlags` than ? Or stay with Filter ? I think both is fine

swozniewski: I think this is the only usecase for the python-filter. That's why I turned the previous Filter into BaseFilter and called this one here Filter.
For simpler Filters without shift like the MET-Filter the simple Producer was sufficient.
But actually we could also use the BaseFilter (or the Filter without subproducers would also work) just to avoid confusion by the naming.

harrypuuter (approved):

PR #69, opened 2021-06-14
Extralepveto

swozniewski:


physicsobjects.hxx
```diff
@@ -107,6 +107,31 @@ auto CombineMasks(auto df, const std::string &maskname,
...
+/// false
+///
+/// \param[in] df the input dataframe
+/// \param[out] outputmaskname the name of the new mask to be added as column to
+/// the dataframe \param[in] inputmaskname the name of the input mask \param[in]
+/// ditaupair name of the column of the ditaupair \param[in] index index of the
+/// tau candidate to be ignored by mask
+///
+/// \return a dataframe containing the new mask
+template <class... Masks>
```
harrypuuter, line 120: this is not used right ?

config/config.py
```diff
@@ -158,14 +169,14 @@ def build_config():
         "tauES_1prong0pizeroUp",
         {"global": {"tau_ES_shift_DM0": 1.002}},
         [[TauPtCorrection, "global"]],
-        [[LVMu1, "mt"]],
+        [[LVMu1, "mt"], [VetoMuons, "mt"]],
```
harrypuuter, line 172: is this the sanitizing, such that the ES shift is not propagated to the vetoMuons ?

Maybe we should also add the argument names here, like we do for producer instances, to make it easier to understand why this is set here.

harrypuuter (changes requested):
two small things, otherwise very nice :+1:

physicsobjects.hxx
swozniewski: oops, indeed. Copy-paste-error

config/config.py
swozniewski: Maybe just for the sanetize_producers argument, which only pops up in some cases? I think repeating all arguments would be too much (e.g. looking at the JEC-shifts-config)

harrypuuter (approved):

config/config.py
harrypuuter: yep, fine with me

PR #70, opened 2021-06-17
add dilepton vetos

### LINKED ISSUE #60: Additional Lepton Veto

swozniewski:
resolves #60


code_generation/producers/muons.py
```diff
@@ -120,3 +120,47 @@
...
+    name="DiMuonVetoIDCut",
+    call='physicsobject::muon::CutID({df}, {output}, "{dimuonveto_id}")',
+    input=[],
+    output=[],
+    scopes=["global"],
+)
+DiMuonVetoMuons = ProducerGroup(
+    name="DiMuonVetoMuons",
+    call="physicsobject::CombineMasks({df}, {output}, {input})",
+    input=MuonEtaCut.output + MuonDxyCut.output + MuonDzCut.output + MuonIsoCut.output,
```
harrypuuter, line 145: i love it !

physicsobjects.hxx
```diff
@@ -203,6 +203,57 @@ auto ObjectMassCorrectionWithPt(auto df, const std::string corrected_mass,
...
+/// \param[in] leptons_phi name of the input phi column of the lepton collection
+/// \param[in] leptons_mass name of the input mass column of the lepton
+/// collection \param[in] leptons_charge name of the input charge column of the
+/// lepton collection \param[in] leptons_mask name of the input mask column of
+/// the lepton collection that marks lepton to be taken into account \param[in]
+/// dR_cut minimum required angular distance between the leptons
+///
+/// \return a dataframe containing the modified object masses
+auto CheckForDiLeptonPairs(auto df, const std::string output_flag,
+                           const std::string leptons_pt,
```
harrypuuter, line 221: add reference here for the inputs

physicsobjects.hxx
```diff
@@ -203,6 +203,57 @@ auto ObjectMassCorrectionWithPt(auto df, const std::string corrected_mass,
     return df1;
 }
 
+/// Function to correct object mass in alignment with object pt correction
```
harrypuuter, line 206: thats not correct right ?

physicsobjects.hxx
```diff
@@ -203,6 +203,57 @@ auto ObjectMassCorrectionWithPt(auto df, const std::string corrected_mass,
...
+             it1 != valid_lepton_indices.end(); it1++) {
+            for (auto it2 = it1 + 1; it2 != valid_lepton_indices.end(); it2++) {
+                if (charge_values.at(*it1) != charge_values.at(*it2)) {
+                    auto p4_1 = ROOT::Math::PtEtaPhiMVector(
+                        pt_values.at(*it1), eta_values.at(*it1),
+                        phi_values.at(*it1), mass_values.at(*it1));
+                    auto p4_2 = ROOT::Math::PtEtaPhiMVector(
+                        pt_values.at(*it2), eta_values.at(*it2),
+                        phi_values.at(*it2), mass_values.at(*it2));
+                    if (ROOT::Math::VectorUtil::DeltaR(p4_1, p4_2) >= dR_cut)
```
harrypuuter, line 244: I was wondering, is this faster or slower compared to a version, where you calculate the deltaR for every iteration and store the bool if the deltaR requirement is passed and then return an any of this vector, so basically running without the if statement here

harrypuuter (changes requested):

physicsobjects.hxx
harrypuuter: in the lines below as well

physicsobjects.hxx
swozniewski: I think in this case the clear advantage is that the loop stops as soon as one suitable pair is found.

swozniewski: Ok, most events don't have such a pair and the full loop runs. But I think that the Any function must do something similar internally. If you compare directly, you save the memory access for the result of the comparison.

PR #71, opened 2021-06-21
prepare era and sample specific configs

swozniewski:


config/config.py
```diff
@@ -112,6 +118,10 @@ def build_config():
         ],
     }
 
+    config["producer_mods"] = [
```
harrypuuter, line 121: I would stick with writing out all names, so 

RemoveProducer instead of Prod_Remove
AppendProducer instead of Prod_Append

harrypuuter: also `producer_modifications` instead of `producer_mods`

config/utility.py
```diff
@@ -13,3 +16,80 @@ def AddSystematicShift(
...
+
+class ProducerRule:
+    def __init__(self, producers, samples, scopes, invert=False, update_output=True):
+        self.producers = producers
+        self.samples = samples
+        self.scopes = scopes
+        self.invert = invert
+        self.update_output = update_output
+
+    def applicable(self, sample):
```
harrypuuter, line 68: what does this function do ?

config/utility.py
```diff
@@ -13,3 +16,80 @@ def AddSystematicShift(
...
+            output.extend(CollectProducerOutput(prod))
+    return output
+
+
+class ProducerRule:
+    def __init__(self, producers, samples, scopes, invert=False, update_output=True):
+        self.producers = producers
+        self.samples = samples
+        self.scopes = scopes
+        self.invert = invert
```
harrypuuter, line 65: What is this parameter needed for ?

config/utility.py
```diff
@@ -13,3 +16,80 @@ def AddSystematicShift(
...
+        self.update_output = update_output
+
+    def applicable(self, sample):
+        applicable = sample in self.samples
+        if self.invert:
+            applicable = not applicable
+        return applicable
+
+    def operate(self, item, item_list):
```
harrypuuter, line 74: so this does nothing yet ?

harrypuuter (changes requested):
I have to say, i nave some trouble understanding, how this is supposed for work. Maybe the code just needs some more documentation ? The era and sample information will be added to the build_config function i guess ? But there is nothing there yet

config/utility.py
swozniewski: inversion of the sample selection

config/utility.py
swozniewski: checks whether modification should be applied based on samples and inversion flag

config/utility.py
swozniewski: only in the derived classes

swozniewski: only for doc that this should be defined in derived classes

harrypuuter (approved):

PR #81, opened 2021-06-29
Htxs

swozniewski:


htxs.hxx
```diff
@@ -0,0 +1,115 @@
...
+        WeightsGraphs[2] =
+            (TGraphErrors *)rootFile.Get("gr_NNLOPSratio_pt_mcatnlo_2jet");
+        WeightsGraphs[3] =
+            (TGraphErrors *)rootFile.Get("gr_NNLOPSratio_pt_mcatnlo_3jet");
+    } else
+        Logger::get("ggHNLLOWeights")
+            ->critical("WARNING: Invalid ggH generator configured. "
+                       "ggHNNLOWeights cannot be determined!");
+    rootFile.Close();
+    const Float_t cutoff[4] = {125.0, 625.0, 800.0, 925.0};
```
harrypuuter, line 51: I guess these values dont have to be changable via the config, so that is fine here

htxs.hxx
```diff
@@ -0,0 +1,115 @@
...
+ */
+auto qqH_WG1_uncertainties(auto &df,
+                           const std::vector<std::string> &weight_names,
+                           const std::string &htxs_flag,
+                           const size_t &idx = 0) {
+    if (idx >= weight_names.size()) {
+        return df;
+    }
+    auto dflamda = [idx](const int &stxs1flag) {
+        if (stxs1flag >= 200 && stxs1flag < 300) {
```
harrypuuter, line 106: Maybe add a comment in the function documentation, what does happen for these values

htxs.hxx
harrypuuter: but you could add this in the documentation of the function

htxs.hxx
```diff
@@ -0,0 +1,115 @@
...
+ * module.
+ * @returns a dataframe with the weight column included
+ */
+auto ggH_WG1_uncertainties(auto &df,
+                           const std::vector<std::string> &weight_names,
+                           const std::string &htxs_flag,
+                           const std::string &htxs_pth,
+                           const std::string &htxs_njets) {
+    auto df1 = df.Define(
+        "ggH_WG1_uncertainties",
```
harrypuuter, line 79: This will cause problems if we run this function within the channel scopes, since the same quantity would be used for the e.g. et and mt. Either we set this string in the python part, or the stxs stuff can only run in the global scope.

htxs.hxx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * @brief Function to derive the WG1 ggH uncertainties
+ *
+ * @param weight_names Names of the derived weight in the dataframe in the order
+ * given by the WG1 macro.
+ * @param htxs_flag Name of the column with the fine htxs stage1.1 flag.
+ * module.
+ * @returns a dataframe with the weight column included
```
harrypuuter, line 96: This documentation does not match the function. The idx is not mentioned, also that the function may be called iteratively should be mentioned here

harrypuuter (changes requested):

code_generation/producers/event.py
```diff
@@ -46,3 +46,54 @@
     scopes=["global"],
     subproducers=[DiElectronVeto, DiMuonVeto],
 )
+
+GGH_NNLO_Reweighting = Producer(
+    name="GGH_NNLO_Reweighting",
+    call='htxs::ggHNLLOWeights({df}, {output}, "{ggHNNLOweightsRootfile}", "{ggH_generator}", {input})',
+    input=[nanoAOD.HTXS_Higgs_pt, nanoAOD.HTXS_njets30],
+    output=[q.ggh_NNLO_weight],
+    scopes=["global", "em", "et", "mt", "tt"],
```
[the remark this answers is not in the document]
swozniewski: right, I think it can either be used in global or in all other channels in case one wants to have certain filters run first. Therefore, I'd allow everything here.

htxs.hxx
```diff
@@ -0,0 +1,115 @@
...
+ */
+auto ggHNLLOWeights(auto &df, const std::string &weight_name,
+                    const std::string &rootfilename,
+                    const std::string &generator, const std::string &htxs_pth,
+                    const std::string &htxs_njets) {
+    TGraphErrors *WeightsGraphs[4];
+    TFile rootFile(rootfilename.c_str(), "READ");
+    if (generator == "powheg") {
```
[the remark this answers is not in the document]
swozniewski: this is run once. Wouldn't an enum be a bit overdoing it?

htxs.hxx
swozniewski: ok

htxs.hxx
harrypuuter: ah no, i am wrong, we have different frames for each scope right ?

swozniewski: right, should be no problem

htxs.hxx
swozniewski: ah, that's maybe some unchanged copy-paste

htxs.hxx
harrypuuter: ok, its fine then

harrypuuter (approved):

PR #85, opened 2021-07-02
MET

### LINKED ISSUE #58: Add MET things
For now, PUPPIMET only

- [x] Readout of MET
- [x] Readout of METCoV
- [x] Propagation of Corrections for 
	- [x] Jets, 
	- [x] electrons, 
	- [x] taus,
	- [x] muons (?)
- [x] Readout of MetUncertainties --> MetUnclustered
- [x] MET Recoil Corrections (special for WJets)
- [x] Calculation of MET related quantities ala `mt_tot`

harrypuuter:
Resolves #58


harrypuuter:
In the past, we had some METResolution added to the final MET Value after Recoil Corrections:

https://github.com/KIT-CMS/KITHiggsToTauTau/blob/33985dfaec23207312c05bf6e910006160950e59/interface/Producers/MetCorrectors.h#L446

This i did not include yet

code_generation/producer.py
```diff
@@ -129,13 +129,28 @@ def writecall(self, config, scope, shift=""):
...
+                and config[shift][scope][para]
+            ):
+                log.debug("Found a boolean True ! - converting to C++ syntax")
+                config[shift][scope][para] = "true"
+            if (
+                isinstance(config[shift][scope][para], bool)
+                and not config[shift][scope][para]
+            ):
+                log.debug("Found a boolean False ! - converting to C++ syntax")
+                config[shift][scope][para] = "false"
```
swozniewski, line 146: It would be a bit shorter and I think a bit better understandible if you had the instance checked first and then the decision true/false inside.

code_generation/quantities/nanoAOD.py
```diff
@@ -72,3 +73,26 @@
...
+## TODO Swich to Puppi versions for METCOV and Signifiance as soon as they are in the nanoAOD
+MET_covXX = NanoAODQuantity("MET_covXX")
+MET_covXY = NanoAODQuantity("MET_covXY")
+MET_covYY = NanoAODQuantity("MET_covYY")
+MET_significance = NanoAODQuantity("MET_significance")
+
+MET_phi = NanoAODQuantity("PuppiMET_phi")
+MET_pt = NanoAODQuantity("PuppiMET_pt")
+MET_sumEt = NanoAODQuantity("PuppiMET_sumEt")
+MET_pt.register_external_shift(
```
swozniewski, line 87: I think this should be done in the config because analysers may want to include it or maybe not

code_generation/quantity.py
```diff
@@ -1,11 +1,14 @@
 import logging
+import time
```
swozniewski, line 2: why?

code_generation/quantity.py
```diff
@@ -1,11 +1,14 @@
...
 
 log = logging.getLogger(__name__)
 
 
 class Quantity:
     def __init__(self, name):
         self.name = name
+        # structure for storing shifts:
+        # {scope1: {shift1 : name2}, {shift2: name2}, scope2: {shift1 : name2}, {shift2: name2}, ...}
```
swozniewski, line 11: better use python dict syntax which was maybe intended anyway (remove brackes between shift1 and shift2)

code_generation/quantity.py
```diff
@@ -26,23 +29,32 @@ def reserve_scope(self, scope):
...
+                leaf = self.shifts["global"][shift]
+        elif scope in self.shifts.keys():
+            if shift in self.shifts[scope].keys():
+                leaf = self.shifts[scope][shift]
+        return leaf
 
     def get_leafs_of_scope(self, scope):
-        return [self.name] + [self.name + shift for shift in self.get_shifts(scope)]
+        result = [self.name] + [
+            self.shifts[scope][shift] for shift in self.get_shifts(scope)
```
swozniewski, line 44: I think this does not work correctly. get_shifts should give you all shifts in global and scope but you try to look them up in self.shifts[scope].
Instead of directly accessing the self.shifts dict you could use get_leaf(...) in the loop.

code_generation/quantity.py
```diff
@@ -129,3 +143,12 @@ def reserve_scope(self, scope):
             )
         )
         raise Exception
+
+    # if the shifted version of a quantity already exists in the input,
+    # this function can be used to register an
+    # branch from the input as a shifted version of a quantity
+    def register_external_shift(self, shift, external_name):
+        shift = "__" + shift
```
swozniewski, line 151: better drop this line and add it in the config utility instead in order to be consistent with the implementation of the other variations

config/config.py
```diff
@@ -264,12 +293,82 @@ def build_config(era, sample):
...
         sanetize_producers=[[LVMu1, "mt"], [VetoMuons, "mt"]],
     )
-    # # Jet energy resolution
-    # shift_dict = {"JE_reso_shift": 1}
-    # AddSystematicShift(config, "jerUncUp", {"global": shift_dict}, [[JetEnergyCorrection, "global"]])
-    # shift_dict = {"JE_reso_shift": -1}
-    # AddSystematicShift(config, "jerUncDown", {"global": shift_dict}, [[JetEnergyCorrection, "global"]])
-    # # Jet energy scale
+    # Add MET shifts
+    AddSystematicShift(
```
swozniewski, line 297: This needs a separate helper function (see comment above) that shifts the nanoAOD quantity (and not the producer which is done here without necessity)

swozniewski (changes requested):
As expected, this is a pretty large PR :) This is part 1 of my review.

config/config.py
harrypuuter: So you would recommend to remove the `register_external_shift` function from the NanoAODQuantities definition file and instead use a new helper function to register these shifts for the quantities ?

code_generation/quantity.py
harrypuuter: wrong here

code_generation/quantities/nanoAOD.py
harrypuuter: this is reworked now :)

code_generation/producer.py
harrypuuter: yeah true, it should be better now ..

harrypuuter:
I also added the calculation of various met related quantities to this PR

code_generation/quantity.py
```diff
@@ -26,39 +38,106 @@ def reserve_scope(self, scope):
...
+            if "global" in self.shifts.keys():
+                # make a copy of the global shifts
+                self.shifts[scope] = self.shifts["global"]
             else:
-                if scope in self.children.keys():
-                    for c in self.children[scope]:
-                        c.shift(name, scope)
+                self.shifts[scope] = {}
+        scopes = [scope]
+        if scope == "global" and scope in self.shifts.keys():
```
swozniewski, line 102: the second part is anyway true

code_generation/quantity.py
```diff
@@ -26,39 +38,106 @@ def reserve_scope(self, scope):
...
+            if name not in self.shifts[scope]:
+                self.shifts[scope][name] = self.name + name
+                if scope == "global":  # shift children in all scopes if scope is global
+                    for any_scope in self.children:
+                        for c in self.children[any_scope]:
+                            c.shift(name, any_scope)
+                else:
+                    if scope in self.children.keys():
+                        for c in self.children[scope]:
+                            c.shift(name, scope)
```
swozniewski, line 115: This should not be done within the loop for each scope but only for the initial one.
After moving it outside the loop, mind the conflicting variable name `scope` inside and outside the loop.

code_generation/quantity.py
```diff
@@ -26,39 +38,106 @@ def reserve_scope(self, scope):
...
+        Returns:
+            None
+        """
         log.debug("Make quantity {} ignore shift {}".format(self.name, name))
-        if not scope in self.ignored_shifts.keys():
-            self.ignored_shifts[scope] = set()
-        self.ignored_shifts[scope].add(name)
+        if scope not in self.ignored_shifts.keys():
+            self.ignored_shifts[scope] = {}
+        self.ignored_shifts[scope] = name
```
swozniewski, line 130: this looks wrong. The name replaces the dict instead of filling it. Why do you want to replace the set by a dict?

code_generation/quantity.py
```diff
@@ -129,3 +219,15 @@ def reserve_scope(self, scope):
...
+    # this function can be used to register an
+    # branch from the input as a shifted version of a quantity
+    def register_external_shift(self, shift, scope, external_name):
+        if scope not in self.shifts.keys():
+            self.shifts[scope] = {}
+        if scope == "global":
+            for allscope in self.shifts.keys():
+                self.shifts[allscope][shift] = external_name
+        else:
+            self.shifts[scope][shift] = external_name
```
swozniewski, line 233: I'd put everything into global and remove the method argument `scope`. There is no need for different scopes in nanoAOD quantities

swozniewski: Furthermore, you need to take care of shifting the children via the usual shift-function

config/utility.py
```diff
@@ -18,6 +18,20 @@ def AddSystematicShift(
...
+    config, name, base_producers, external_dict, sanetize_producers=[]
+):
+    for scope in external_dict.keys():
+        for quantity in external_dict[scope].keys():
+            quantity.register_external_shift(
+                shift="__" + name,
+                scope=scope,
+                external_name=external_dict[scope][quantity],
+            )
+    AddSystematicShift(config, name, {}, base_producers, sanetize_producers)
```
swozniewski, line 32: I think, there is no need for the call of `AddSystematicShift`. Also, there is no need for scope treatment (see comment above). You only need to call register_external_shift with shiftname and quantityname. If function above is adapted correspondingly this will propagate through all dependent quantities. This is independent from producers, which were only used beforehand to create dependencies between quantities.
Sidemark: In the AddSystematicShift method, the producers are used because systematic variations are generated by changing their parameters and behavior such that there is a direct logic link between variation and producer (this is basically userfriendliness). For the propagation it would also work to directly shift their output quantities.

swozniewski (changes requested):
ok, this is another iteration over the python part. Did not have a look yet at the cpp part.

swozniewski (changes requested):
another part of review. not yet complete

code_generation/quantity.py
harrypuuter: True

code_generation/quantity.py
harrypuuter: I tried this, but i think this does not work.. 

When the writecall function is run, `get_leaf(shift, scope)` is run. Since the NanoAODQuantity only exists in global, the `BuildMetVector` function, which currently sits in the `mt` scope does not get the shifted version

src/RecoilCorrections/METSystematics.hxx
```diff
@@ -0,0 +1,57 @@
...
+  public:
+    METSystematic(std::string filepath);
+    ~METSystematic(){};
+
+    void ApplyMETSystematic(float metPx, float metPy, float genVPx,
+                            float genVPy, float visVPx, float visVPy, int njets,
+                            int sysType, int shiftType, float &metShiftPx,
+                            float &metShiftPy);
+
+    void ShiftMEt(float metPx, float metPy, float genVPx, float genVPy,
```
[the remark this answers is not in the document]
harrypuuter: fixed

code_generation/quantity.py
swozniewski: I'd assume that this is because the shift was not correctly propagated to the children (second point above). Check how it is done in the shift function ( https://github.com/KIT-CMS/CROWN/blob/main/code_generation/quantity.py#L52-L55 ). If the scope is global, children within all scopes must be shifted with the respective scope.

src/RecoilCorrections/METSystematics.cxx
```diff
@@ -0,0 +1,229 @@
+#include "METSystematics.hxx"
+#include "../utility/Logger.hxx"
+
+METSystematic::METSystematic(std::string filepath) {
+
+    fileName = filepath;
+    TFile *file = new TFile(fileName);
+    if (file->IsZombie()) {
+        Logger::get("RecoilCorrector")
```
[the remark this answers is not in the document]
harrypuuter: fixed

src/RecoilCorrections/METSystematics.cxx
```diff
@@ -0,0 +1,229 @@
...
+            ->debug("Histogram nJetBinsH should be contained in file {}",
+                    fileName);
+        Logger::get("METSystematics")
+            ->debug("Check content of the file {}", fileName);
+        exit(-1);
+    }
+
+    nJetBins = jetBinsH->GetNbinsX();
+    std::vector<TString> JetBins;
+    JetBins.clear();
```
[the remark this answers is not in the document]
harrypuuter: removed

code_generation/quantity.py
harrypuuter: For the children, it does work, only for the variable itself, it does not

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+                                 const ROOT::RVec<float> &jet_pt,
+                                 const ROOT::RVec<float> &jet_eta,
+                                 const ROOT::RVec<float> &jet_phi,
+                                 const ROOT::RVec<float> &jet_mass) {
+        ROOT::Math::PtEtaPhiMVector corrected_met;
+        ROOT::Math::PtEtaPhiMVector uncorrected_jet;
+        ROOT::Math::PtEtaPhiMVector corrected_jet;
+        float corr_x = 0.0;
+        float corr_y = 0.0;
+        // now loop though all jets in the event
```
swozniewski, line 180: ->through

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+        // component of the MET according to the correction of the lepton
+        // Recalculate MET with corrected lepton energies :
+        // MetX_corrected = MetX + Px - Px_corrected
+        // MetY_corrected = MetY + Py - Py_corrected
+        // MET_corrected = sqrt(MetX_corrected * MetX_corrected + MetY_corrected
+        // * MetY_corrected)
+        float corr_x = uncorrected_object.Px() - corrected_object.Px();
+        float corr_y = uncorrected_object.Py() - corrected_object.Py();
+        float MetX = met.Px() - corr_x;
+        float MetY = met.Py() - corr_y;
```
swozniewski, line 87: I think your formula in the comments above is correct, but then you should have + here, given the way the correction is built in the two lines above.

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+                corrected_jet = ROOT::Math::PtEtaPhiMVector(
+                    jet_pt.at(index), jet_eta.at(index), jet_phi.at(index),
+                    jet_mass.at(index));
+                // update the correction factors that are applied to the met
+                corr_x += uncorrected_jet.Px() - corrected_jet.Px();
+                corr_y += uncorrected_jet.Py() - corrected_jet.Py();
+            }
+        }
+        float MetX = met.Px() - corr_x;
+        float MetY = met.Py() - corr_y;
```
swozniewski, line 198: also here, this should probably be +

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+-------------------------------------|-------|-------
+ isPrompt                            | 1     | 0
+ isDecayedLeptonHadron               | 2     | 1
+ isTauDecayProduct                   | 4     | 2
+ isPromptTauDecayProduct             | 8     | 3
+ isDirectTauDecayProduct             | 16    | 4
+ isDirectPromptTauDecayProduct       | 32    | 5
+ isDirectHadronDecayProduct          | 64    | 6
+ isHardProcess                       | 128   | 7
+ fromHardProcess                     | 265   | 8
```
swozniewski, line 239: 256

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+        auto RecoilCorrections = [sysType, systematics, shiftType, corrector](
+                                     ROOT::Math::PtEtaPhiMVector &met,
+                                     const ROOT::RVec<float> &genparticle_pt,
+                                     const ROOT::RVec<float> &genparticle_eta,
+                                     const ROOT::RVec<float> &genparticle_phi,
+                                     const ROOT::RVec<float> &genparticle_mass,
+                                     const ROOT::RVec<int> &genparticle_id,
+                                     const ROOT::RVec<int> &genparticle_status,
+                                     const ROOT::RVec<float> &jet_pt) {
+            // TODO is this the correct number of jets ?
```
swozniewski, line 301: I think we need some extra treatment for WJets

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+            float visPx = 0.; // visible (generator) Z(W) px
+            float visPy = 0.; // visible (generator) Z(W) py
+            // now loop though all genparticles in the event
+            for (std::size_t index = 0; index < genparticle_id.size();
+                 ++index) {
+                // consider a genparticle,
+                // 1. if it is a lepton and fromHardProcess --> bit 8 from
+                // status
+                // 2. if it is isDirectHardProcessTauDecayProduct --> bit 10
+                // in status
```
swozniewski, line 322: We will need the same procedure of building the gen V for the Z pt mass reweighting. We should think about putting this into an extra function, maybe even an extra producer and df column.

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+            Logger::get("RecoilCorrections")->debug("MetX {} ", MetX);
+            Logger::get("RecoilCorrections")->debug("MetY {} ", MetY);
+            Logger::get("RecoilCorrections")
+                ->debug("correctedMetX {} ", correctedMetX);
+            Logger::get("RecoilCorrections")
+                ->debug("correctedMetY {} ", correctedMetY);
+            Logger::get("RecoilCorrections")->debug("old met {} ", met.Pt());
+            corrector->CorrectWithHist(MetX, MetY, genPx, genPy, visPx, visPy,
+                                       nJets30, correctedMetX, correctedMetY);
+            // only apply shifts is the correpsonding variables are set
```
swozniewski, line 361: is -> if

swozniewski (commented):
almost done with the review

src/quantities.hxx
```diff
@@ -177,14 +166,220 @@ auto m_vis(auto &df, const std::string &outputname,
...
+ * @return a new dataframe with the new column
+ */
+auto mTdileptonMET(auto &df, const std::string &outputname,
+                   const std::string &p_1_p4, const std::string &p_2_p4,
+                   const std::string &met) {
+    auto calculate_mTdileptonMET = [](ROOT::Math::PtEtaPhiMVector &p_1_p4,
+                                      ROOT::Math::PtEtaPhiMVector &p_2_p4,
+                                      ROOT::Math::PtEtaPhiMVector &met) {
+        auto dileptonsystem = p_1_p4 + p_2_p4;
+        return vectoroperations::calculateMT(dileptonsystem, met);
```
swozniewski, line 265: better write the sum directly into the function call?

src/quantities.hxx
```diff
@@ -177,14 +166,220 @@ auto m_vis(auto &df, const std::string &outputname,
...
+ * @return a new dataframe with the new column
+ */
+
+auto pt_tt(auto &df, const std::string &outputname, const std::string &p_1_p4,
+           const std::string &p_2_p4, const std::string &met) {
+    auto calculate_pt_tt = [](ROOT::Math::PtEtaPhiMVector &p_1_p4,
+                              ROOT::Math::PtEtaPhiMVector &p_2_p4,
+                              ROOT::Math::PtEtaPhiMVector &met) {
+        auto dileptonmet = p_1_p4 + p_2_p4 + met;
+        return (float)dileptonmet.Pt();
```
swozniewski, line 308: could be done in one line

src/quantities.hxx
```diff
@@ -177,14 +166,220 @@ auto m_vis(auto &df, const std::string &outputname,
...
+             const std::string &jet_2_p4, const std::string &met) {
+    auto calculate_pt_ttjj = [](ROOT::Math::PtEtaPhiMVector &p_1_p4,
+                                ROOT::Math::PtEtaPhiMVector &p_2_p4,
+                                ROOT::Math::PtEtaPhiMVector &jet_1_p4,
+                                ROOT::Math::PtEtaPhiMVector &jet_2_p4,
+                                ROOT::Math::PtEtaPhiMVector &met) {
+        if (jet_1_p4.pt() < 0.0 || jet_2_p4.pt() < 0.0)
+            return default_float;
+        auto jetlepmet = p_1_p4 + p_2_p4 + met + jet_1_p4 + jet_2_p4;
+        return (float)jetlepmet.Pt();
```
swozniewski, line 339: same here, but ok, at some point the line grows a bit large...

src/quantities.hxx
```diff
@@ -177,14 +166,220 @@ auto m_vis(auto &df, const std::string &outputname,
...
+
+auto mt_tot(auto &df, const std::string &outputname, const std::string &p_1_p4,
+            const std::string &p_2_p4, const std::string &met) {
+    auto calculate_mt_tot = [](ROOT::Math::PtEtaPhiMVector &p_1_p4,
+                               ROOT::Math::PtEtaPhiMVector &p_2_p4,
+                               ROOT::Math::PtEtaPhiMVector &met) {
+        const float mt_1 = vectoroperations::calculateMT(p_1_p4, met);
+        const float mt_2 = vectoroperations::calculateMT(p_2_p4, met);
+        const float mt_mix = vectoroperations::calculateMT(p_1_p4, p_2_p4);
+        return (float)sqrt(mt_1 * mt_1 + mt_2 * mt_2 + mt_mix * mt_mix);
```
swozniewski, line 369: which doesn't seem to bother you here :)

code_generation/quantity.py
```diff
@@ -6,16 +6,25 @@
...
         self.shifts = {}
         self.ignored_shifts = {}
         self.children = {}
         self.defined_for_scopes = []
         log.debug("Setting up new Quantity {}".format(self.name))
 
-    # the scopes, in which a quantity is used as an output is tracked in the output_scopes list.
-    # This check is triggered for every producer.
-    # If a quantity is already used within a given scope as output, this will result in an exception.
+    #
```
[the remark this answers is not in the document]
harrypuuter: quantity class changes are reversed

code_generation/quantity.py
```diff
@@ -26,39 +35,106 @@ def reserve_scope(self, scope):
...
+            scope (str): Scope for which the leaf should be returned
+        Returns:
+            str. Name of the leaf
+        """
+        log.debug("{} - getting shift {} for scope {}".format(self.name, shift, scope))
+        leaf = self.name
+        if scope in self.shifts.keys():
+            if shift in self.shifts[scope].keys():
+                leaf = self.shifts[scope][shift]
+        return leaf
```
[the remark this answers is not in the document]
harrypuuter: quantity class changes are reversed

code_generation/quantity.py
harrypuuter: quantity class changes are reversed

code_generation/quantity.py
harrypuuter: quantity class changes are reversed

code_generation/quantity.py
harrypuuter: quantity class changes are reversed

config/utility.py
harrypuuter: quantity class changes are reversed

src/RecoilCorrections/RecoilCorrector.hxx
```diff
@@ -0,0 +1,127 @@
...
+        return binN;
+    }
+
+    TString fileName;
+
+    void InitMEtWeights(TFile *file, TString _perpZStr, TString _paralZStr,
+                        int nZPtBins, float *ZPtBins, TString *_ZPtStr,
+                        int nJetsBins, TString *_nJetsStr);
+
+    void InitMEtWeights(TFile *file, const std::vector<float> &ZPtBins,
```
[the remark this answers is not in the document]
harrypuuter: Changed

src/met.hxx
```diff
@@ -0,0 +1,389 @@
...
+ * @param outputname name of the new column containing the met phi value
+ * @return a new dataframe containing the met phi value
+ */
+auto metPhi(auto df, const std::string &metvector,
+            const std::string &outputname) {
+    auto GetMetPhi = [](const ROOT::Math::PtEtaPhiMVector met) {
+        return met.Phi();
+    };
+    return df.Define(outputname, GetMetPhi, {metvector});
+}
```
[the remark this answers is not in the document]
harrypuuter: aaah yes, good catch, initially i was using `PtEtaPhiEVector` for the met, which made it necessary to write seperate functions, but since it is possible to cast a `PtEtaPhiEVector` to a `PtEtaPhiMVector` this is not needed anymore !

src/vectoroperations.hxx
```diff
@@ -0,0 +1,24 @@
...
+ missing transverse energy.
+ *
+ * @param particle lorentz vector of the particle
+ * @param met lorentz vector of the missing transverse energy
+ * @return the transverse mass of the particle
+ */
+auto calculateMT(ROOT::Math::PtEtaPhiMVector &particle,
+                 ROOT::Math::PtEtaPhiMVector &met) {
+    return (float)sqrt(2 * particle.Pt() * met.Pt() *
+                       (1. - cos(particle.Phi() - met.Phi())));
```
swozniewski, line 22: I think with the cos, this subtraction of phi values works fine. But it's a bit delicate, so just to make sure that you thought about it as well.

src/met.hxx
harrypuuter: do we run these omly for WJets or also for other samples ?

For WJets we just add one jet here so 

```
int nJets30 = jets_above30.size() + 1;
```

src/met.hxx
harrypuuter: good catch !

src/met.hxx
harrypuuter: hmm shot, i even had it as an extra function before

src/met.hxx
swozniewski: also for DY and Signal afaik

src/quantities.hxx
harrypuuter: This is the way i thought it would be best readable :)

src/quantities.hxx
harrypuuter: it did !

src/quantities.hxx
harrypuuter: changed

src/met.hxx
harrypuuter: ok, then we will need an additional flag for this

harrypuuter: Is that possible, or do we need two separate producers in the python part ?

swozniewski: I think a config parameter which is steered via the sample key would do the job.

src/met.hxx
harrypuuter: done

src/met.hxx
harrypuuter: done

code_generation/producers/met.py
```diff
@@ -0,0 +1,148 @@
...
+    subproducers=[
+        BuildMetVector,
+        MetCov00,
+        MetCov01,
+        MetCov10,
+        MetCov11,
+        MetSumEt,
+        PropagateLeptonsToMet,
+        PropagateJetsToMet,
+        calculateGenBosonVector,
```
swozniewski, line 143: Let's stick to the convention to begin producer names with capital letters

code_generation/quantity.py
```diff
@@ -12,12 +12,24 @@ def __init__(self, name):
...
+        Function to reserve a scope for a given quantity. The scopes, in which a quantity is used
+        as an output are tracked in the output_scopes list.
+        If a quantity is already used within a given scope as output, this will result in an exception.
+        This check is triggered for every Producer.
+        """
         log.debug("Checking {} / scope {}".format(self.name, scope))
-        if scope not in self.defined_for_scopes:
+        if (scope == "global" and self.defined_for_scopes == []) or (
+            scope not in self.defined_for_scopes
+            and "global" not in self.defined_for_scopes
```
swozniewski, line 31: this line must be scope != "global". Otherwise global could be added via this part of the or even if defined_for_scopes is not empty and the first part fails.

config/utility.py
```diff
@@ -19,13 +19,26 @@ def AddSystematicShift(
         producer[0].shift(name, producer[1])
 
 
+# Function for introducing systematic variations to producers and depending quantities by adding an already shifted input quantity
+def SystematicShiftByInputQuantity(config, shiftname, base_producers, external_dict):
```
swozniewski, line 23: remove base_producers

src/met.hxx
```diff
@@ -0,0 +1,413 @@
...
+                }
+            }
+            ROOT::Math::PtEtaPhiMVector genBoson;
+            ROOT::Math::PtEtaPhiMVector visgenBoson;
+
+            genBoson.SetPxPyPzE(genPx, genPy, 0,
+                                std::sqrt(genPx * genPx + genPy * genPy));
+            visgenBoson.SetPxPyPzE(visPx, visPy, 0,
+                                   std::sqrt(visPx * visPx + visPy * visPy));
+            std::pair<ROOT::Math::PtEtaPhiMVector, ROOT::Math::PtEtaPhiMVector>
```
swozniewski, line 112: can we build the complete 4-vector of the genboson including z-component? we need it to determine the mass for the z-pt-mass-reweighting

swozniewski (changes requested):
Ok great. I still found some problems but it's not much.

src/met.hxx
harrypuuter: added

config/utility.py
harrypuuter: changed

code_generation/quantity.py
harrypuuter: changed

code_generation/producers/met.py
harrypuuter: changed

swozniewski (approved):

PR #93, opened 2021-07-14
Add optimizer to automatically determine a correct producer ordering

### LINKED ISSUE #91: [Request] Function to attempt a correct ordering of producers
**Is your feature request related to a problem? Please describe.**
Currently, it is possible to use an incorrect ordering of producers in the python config. This will result in a runtime error, since there is no check, if the order set in the config is actually valid. 

**Describe the solution you'd like**
Add a new function, something like `ResolveProducerOrdering` and attempt to find a matching producer order there. If no valid ordering can be created, this should result in an error during the config step

harrypuuter:
Resolves #91 

The optimization moves filters to the beginning of the producer ordering and then attempts to solve dependency issues. The optimizer is also able to check, if some inputs cannot be found in the produced outputs and will raise an Exception is this is the case. 

I tested the algorithm with shuffled producer lists for both the global and mt scope and the algorithm worked every time.


code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+        for producer in self.ordering:
+            temp_outputs = producer.get_outputs(self.scope)
+            if temp_outputs != None:
+                outputs.extend(
+                    [
+                        quantity
+                        for quantity in temp_outputs
+                        if not isinstance(quantity, NanoAODQuantity)
+                    ]
+                )
```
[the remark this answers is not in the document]
swozniewski: also see comment above on the NanoAODQuantitiy check

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+                            producer.get_outputs(self.scope),
+                            self.ordering.index(producer),
+                        )
+                    )
+                    producers_to_relocate.add(producer)
+            if self.scope != "global":
+                for producer in self.global_producers:
+                    if input in producer.get_outputs("global"):
+                        found = True
+                        log.debug("found {} in global scope ..".format(input))
```
swozniewski, line 249: Is there a use case for this? Such variables should not be identified as 'wrong', right?

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+                log.error("Please check, if all needed producers are activated")
+                raise Exception
+            wrongProducer, wrong_inputs = self.check_ordering()
+            producers_to_relocate = self.find_inputs(wrongProducer, wrong_inputs)
+            if len(producers_to_relocate) == 0:
+                self.optimized = True
+            else:
+                for producer_to_relocate in producers_to_relocate:
+                    counter += 1
+                    producerIndex = self.get_position(wrongProducer)
```
swozniewski, line 148: better use this function in place below as done for producer_to_relocate

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+    Returns:
+        producers_to_relocate: A list of producers, that have to be run before the given producer
+    """
+
+    def find_inputs(self, producer, inputs):
+        producers_to_relocate = set()
+        log.debug("Trying to find inputs {}".format(inputs))
+        for input in inputs:
+            found = False
+            for producer in self.ordering:
```
swozniewski, line 233: `producer` already used as variable outside the loop. Better use different variable name.

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+        updated_ordering = list(range(self.size))
+        if old_position > new_position:
+            for position in updated_ordering:
+                if position <= old_position and position > new_position:
+                    updated_ordering[position] = position - 1
+                if position == new_position:
+                    updated_ordering[position] = old_position
+        if old_position < new_position:
+            for position in updated_ordering:
+                if position < old_position and position >= new_position:
```
swozniewski, line 280: this is never true. probably mixed up old and new

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+                    updated_ordering[position] = position - 1
```
swozniewski, line 281: should be +1?

swozniewski (changes requested):

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+        self.global_producers = config["producers"]["global"]
+        self.ordering = config["producers"][scope]
+        self.size = len(self.ordering)
+        self.config = config
+        self.scope = scope
+        self.optimized = False
+        self.global_outputs = self.get_global_outputs()
+
+    def __call__(self):
+        return self.producer_list
```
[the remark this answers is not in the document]
harrypuuter: removed

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def get_global_outputs(self):
+        outputs = []
+        for producer in self.global_producers:
+            if producer.get_outputs("global") != None:
+                outputs.extend(
+                    [
+                        quantity
+                        for quantity in producer.get_outputs("global")
+                        if not isinstance(quantity, NanoAODQuantity)
```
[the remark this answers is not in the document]
harrypuuter: i just wanted NanoAOD Quantities to never be considered during the optimization, same as below

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+        self.global_outputs = self.get_global_outputs()
+
+    def __call__(self):
+        return self.producer_list
+
+    """
+    Helper Function to get the position of a producer in the ordering list
+
+    Args:
+        producer: The producer to get the position of
```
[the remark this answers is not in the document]
harrypuuter: yip

code_generation/optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+            else:
+                new_ordering.append(producer)
+        for i, prod in enumerate(self.ordering):
+            log.debug(" --> {}. : {}".format(i, prod))
+        for i, prod in enumerate(new_ordering):
+            log.debug(" --> {}. : {}".format(i, prod))
+        self.ordering = new_ordering
+
+    """
+    The main function of this class. During the optimization, a
```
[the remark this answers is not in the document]
harrypuuter: :+1:

code_generation/optimizer.py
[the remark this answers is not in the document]
harrypuuter: you mean move the output extension after the check ? It has to be done before the check, since for producerGroups, it can very well be the case, that outputs and inputs are created during a single group. 

The optimization is only done on the level of what is provided in the config, so if a producer group is set here, the structure within the group itself is not checked.

code_generation/optimizer.py
harrypuuter: yeah, they are not. This is just for the check, if a input can be found within the given scope or the global scope. If this is not the case, there will be a runtime error later in the c++ execution, so this has to be checked

code_generation/optimizer.py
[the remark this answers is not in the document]
swozniewski: ok, good point!

code_generation/optimizer.py
harrypuuter: true !

code_generation/optimizer.py
harrypuuter: Changed

code_generation/optimizer.py
harrypuuter: yeah that is true. 

This part is mainly implemented for completeness. The way the algorithm is implemented now, this case should never happen, we always want to put producers from a higher position to a lower position.

code_generation/optimizer.py
harrypuuter: jip

code_generation/optimizer.py
swozniewski: ah ok, if an input is not produced at all it will be marked as wrong, too, and end up here

harrypuuter: yes

swozniewski (approved):
very good!

PR #98, opened 2021-07-20
add zptmass reweighting

swozniewski:


src/reweighting.hxx
```diff
@@ -92,4 +93,52 @@ auto topptreweighting(auto &df, const std::string &weightname,
...
+ * @return a new dataframe containing the new column
+ */
+auto zPtMassReweighting(auto &df, const std::string &weightname,
+                        const std::string &gen_boson,
+                        const std::string &workspace_file,
+                        const std::string &functor_name,
+                        const std::string &argset) {
+
+    // retrieve pt and mass of gen boson
+    auto df1 = df.Define(gen_boson + "_pt",
```
harrypuuter, line 117: I would be in favour of putting this and the define below into seperate functions. This way it would also be possible to writeout genboson pt and gensobon mass as quantities.

harrypuuter (changes requested):

src/reweighting.hxx
```diff
@@ -92,4 +93,54 @@ auto topptreweighting(auto &df, const std::string &weightname,
...
+    // retrieve pt and mass of gen boson reconstructed with the method used by
+    // recoil corrections; resulting quantities are only for the purpose of this
+    // method
+    auto df1 = df.Define(gen_boson + "_pt",
+                         [](const std::pair<ROOT::Math::PtEtaPhiMVector,
+                                            ROOT::Math::PtEtaPhiMVector> &p4) {
+                             return (float)p4.first.pt();
+                         },
+                         {gen_boson});
+    auto df2 = df1.Define(gen_boson + "_mass",
```
harrypuuter, line 125: maybe add `internal` `zpt_` or something to the name, so it wont block the `genboson_pt` column for the version where the genboson is read out directly

harrypuuter (changes requested):

harrypuuter (approved):

PR #111, opened 2021-10-12
Rewrite Configuration

harrypuuter:
This is a rather large PR, which concerns a rewrite of the python configuration part of CROWN:

TDR: Instead of using a dict which contains all needed information, a configuration object is introduced, and adding new producers, quantities, shifts etc. is done with defined functions, This should result in much less configuration errors compared to before, where a defined version of a python dictionary was required. 

Additionally, type annotations were added to add functions of the project. To make use of this feature, you should add some type checking tool to your editor, such as pylance (https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) or mypy and activate tpye checking mode.


harrypuuter:
The PR should now be ready for review. The error in the CI is due to a feature, which depends on ROOT 6.25, which is not yet released and therefore no docker container for the CI is available

code_generation/rules.py
```diff
@@ -0,0 +1,177 @@
...
+        self.samples = samples
+        if isinstance(scopes, str):
+            scopes = [scopes]
+        self.scopes = scopes
+        self.invert = invert
+        self.update_output = update_output
+        self.global_scope = "global"
+
+    def set_scopes(self, scopes: List[str]) -> None:
+        if isinstance(scopes, str):
```
rfvc, line 39: Considering the type of `scopes`, this should never be true?

code_generation/modifiers.py
```diff
@@ -0,0 +1,76 @@
...
+class SampleModifier(Modifier):
+    def __init__(
+        self,
+        modifier_dict: ModifierDict,
+        default: Union[str, int, float, bool, None] = None,
+    ):
+        super(SampleModifier, self).__init__(modifier_dict)
+        self.modifier_dict = modifier_dict
+        self.default = default
+        self.samples: List[str] = list(self.modifier_dict.keys())
```
nshadskiy, line 48: Why is this handled differently compared to the EraModifier ?
`self.eras = list(self.modifier_dict.keys())`

code_generation/configuration.py
```diff
@@ -0,0 +1,526 @@
...
+
+        Args:
+            None
+
+        Returns:
+            None
+        """
+        missing_channels = self.channels - self.available_channels
+        if len(missing_channels) > 0:
+            raise ChannelConfigurationError(self.channels, missing_channels)
```
nshadskiy, line 110: Shouldn't this be 
`raise ChannelConfigurationError(missing_channels, self.available_channels)`
based on the definition [here](https://github.com/KIT-CMS/CROWN/blob/eff8756bcf070769078e2c7ed6d840eaac37f1e1/code_generation/exceptions.py#L34) ?

code_generation/producer.py
```diff
@@ -1,16 +1,28 @@
...
         return "{" + key + "}"
 
 
 class Producer:
-    def __init__(self, name, call, input, output, scopes):
+    def __init__(
+        self,
+        name: str,
+        call: str,
+        input: Union[List[q.Quantity], Dict[str, List[q.Quantity]]],
```
nshadskiy, line 22: The input to a Producer can also be from nanoAOD, so can `q.Quantity` and `q.NanoAODQuantity` be seen as the same? Or is this even relevant here?

harrypuuter: A `NanoAODQuantity` is a `Quantity` since  the `NanoAODQuantity`class inherits from `Quantity`, so `q.Quantity` covers both cases

code_generation/configuration.py
harrypuuter: true !

code_generation/rules.py
harrypuuter: yes, it is more of a check for users without type annotation checks

code_generation/configuration.py
```diff
@@ -0,0 +1,526 @@
...
+
+        Args:
+            shift: The shift to be added. This must be a SystematicShift object.
+            samples: The samples to which the shift should be applied. This can be a list of samples or a single sample.
+                If ths option is not set, the shift is applied, regardless of the sample type.
+
+        Returns:
+            None
+        """
+        if not isinstance(shift, SystematicShift):
```
nshadskiy, line 263: This means `SystematicShiftByQuantity` can't  be used yet?

harrypuuter: Same as for the Quantitiy, the `SystematicShiftByQuantity` class inherits from `SystematicShift`so the isInstance call is also true for `SystematicShiftByQuantity`

https://github.com/KIT-CMS/CROWN/blob/eff8756bcf070769078e2c7ed6d840eaac37f1e1/code_generation/systematics.py#L401

code_generation/modifiers.py
harrypuuter: true :+1:

config/config.py
```diff
@@ -336,211 +461,671 @@ def build_config(era, sample, channels, shifts):
...
-    # AddSystematicShift(config, "jecUncAbsoluteDown", {"global": shift_dict}, [[JetEnergyCorrection, "global"]])
+    #########################
+    JEC_sources = '{"SinglePionECAL", "SinglePionHCAL", "AbsoluteMPFBias", "AbsoluteScale", "Fragmentation", "PileUpDataMC", "RelativeFSR", "PileUpPtRef"}'
+    configuration.add_shift(
+        SystematicShift(
+            name="jecUncAbsoluteUp",
+            shift_config={
+                ("et", "mt", "tt", "em", "ee", "mm"): {
+                    "JE_scale_shift": 1,
+                    "JEC_shift_sources": JEC_sources,
```
nshadskiy, line 736: The `JE_scale_shift` and `JEC_shift_sources` are defined in the `"global"` scope. 
https://github.com/KIT-CMS/CROWN/blob/eff8756bcf070769078e2c7ed6d840eaac37f1e1/config/config.py#L117-L118 
Is there no problem to have them now twice in `"global"` and e.g. `"mt"` ?

code_generation/systematics.py
```diff
@@ -0,0 +1,450 @@
...
+            #             if isinstance(temp, list):
+            #                 temp_dict[scope] = temp
+            #             else:
+            #                 errormsg = "Producer {} is not a Producer object or a list of Producer objects".format(
+            #                     temp
+            #                 )
+            #                 raise ValueError(errormsg)
+
+        for scope in scopes:
+            resolved_dict[scope] = []
```
nshadskiy, line 239: I think these two for loops can be merged to one. Or would there be a problem?

code_generation/systematics.py
```diff
@@ -0,0 +1,450 @@
...
+            try:
+                temp = temp_dict[scope]
+                if isinstance(temp, Producer):
+                    resolved_dict[scope] = [temp]
+                else:
+                    resolved_dict[scope] = temp
+            except KeyError:
+                resolved_dict[scope] = []
+                log.debug(
+                    "Setting empy Producer set for shift {} in scope {}".format(
```
nshadskiy, line 249: typo: emp"t"y

code_generation/configuration.py
```diff
@@ -0,0 +1,526 @@
...
+        for scope in scopes_to_test:
+            if (len(self.producers[scope]) == 0) or (
+                scope not in self.channels and scope is not self.global_scope
+            ):
+                log.warning("Removing unrequested / empty scope {}".format(scope))
+                self.scopes.remove(scope)
+                del self.producers[scope]
+                del self.outputs[scope]
+                del self.shifts[scope]
+                del self.config_parameters[scope]
```
nshadskiy, line 347: This might not be really necessary but I would also add 
`del self.available_outputs[scope]`
since it's in the loop
https://github.com/KIT-CMS/CROWN/blob/eff8756bcf070769078e2c7ed6d840eaac37f1e1/code_generation/configuration.py#L177-L182

code_generation/configuration.py
```diff
@@ -0,0 +1,526 @@
...
+        Function used to remove empty scopes from the configuration. This function is called by the optimize function,
+        which should be called after all configuration changes have been made.
+
+        Args:
+            None
+
+        Returns:
+            None
+        """
+        scopes_to_test = [scope for scope in self.scopes]
```
nshadskiy, line 337: Is this necessary since `self.scopes` is already a list
https://github.com/KIT-CMS/CROWN/blob/eff8756bcf070769078e2c7ed6d840eaac37f1e1/code_generation/configuration.py#L90

code_generation/optimizer.py
```diff
@@ -1,77 +1,84 @@
...
     """
 
-    def __init__(self, config, scope):
-        self.global_producers = config["producers"]["global"]
-        self.ordering = config["producers"][scope]
+    def __init__(
+        self,
+        global_producers: List[Producer | ProducerGroup],
+        scope: str,
+        producer_ordering: List[Producer | ProducerGroup],
```
nshadskiy, line 24: `producer_ordering` is missing in the Args description
https://github.com/KIT-CMS/CROWN/blob/847581d22455fcc82e60e06b8752760b5cfc5e91/code_generation/optimizer.py#L29-L32

code_generation/optimizer.py
```diff
@@ -136,42 +148,49 @@ def Optimize(self):
...
-                self.optimized = True
-            else:
-                for producer_to_relocate in producers_to_relocate:
-                    counter += 1
-                    self.relocate_producer(
-                        producer_to_relocate,
-                        self.get_position(producer_to_relocate),
-                        self.get_position(wrongProducer),
-                    )
+            if wrongProducer is not None:
```
nshadskiy, line 151: I think due to this if statement 
https://github.com/KIT-CMS/CROWN/blob/847581d22455fcc82e60e06b8752760b5cfc5e91/code_generation/optimizer.py#L152-L155
this if statement will never be true, because this would only happen if `find_inputs()` finds the input in the "global" scope, but this should already be covered by `check_ordering()`. Am I wrong?

code_generation/producers/pairselection.py
```diff
@@ -102,7 +154,59 @@
...
+    call="lorentzvectors::build({df}, {input_vec}, 0, {output})",
+    input=[
+        q.ditaupair,
+        nanoAOD.Electron_pt,
+        nanoAOD.Electron_eta,
+        nanoAOD.Electron_phi,
+        nanoAOD.Electron_mass,
+    ],
+    output=[q.p4_1_uncorrected],
+    scopes=["mt", "mm"],
```
nshadskiy, line 183: I think this should be
`scopes=["em","et","ee"]`

code_generation/producers/pairselection.py
```diff
@@ -102,7 +154,59 @@
...
+    call="lorentzvectors::build({df}, {input_vec}, 2, {output})",
+    input=[
+        q.ditaupair,
+        nanoAOD.Electron_pt,
+        nanoAOD.Electron_eta,
+        nanoAOD.Electron_phi,
+        nanoAOD.Electron_mass,
+    ],
+    output=[q.p4_2_uncorrected],
+    scopes=["mm"],
```
nshadskiy, line 196: I think this should be
`scopes=["ee"]`

code_generation/producers/genparticles.py
```diff
@@ -127,6 +133,18 @@
     output=[q.gen_match_2],
     scopes=["mt", "et", "tt"],
 )
+gen_taujet_pt_1 = Producer(
+    name="gen_taujet_pt_1",
+    call="quantities::tau::matching_genjet_pt({df}, {output}, 1, {input})",
```
nshadskiy, line 138: I think the index should be "0" here
`call="quantities::tau::matching_genjet_pt({df}, {output}, 0, {input})"`
since it's the first particle.

code_generation/producers/genparticles.py
```diff
@@ -139,42 +157,156 @@
...
+    scopes=["em", "mm"],
+    subproducers=[gen_pt_2, gen_eta_2, gen_phi_2, gen_mass_2, gen_pdgid_2],
+)
+UnrollGenElLV1 = ProducerGroup(
+    name="UnrollGenElLV1",
     call=None,
     input=None,
     output=None,
-    scopes=["mt", "et", "tt", "em"],
+    scopes=["em", "ee" "et"],
```
nshadskiy, line 181: Typo `"ee", "et"` missing comma

code_generation/producers/genparticles.py
harrypuuter: yes

code_generation/optimizer.py
harrypuuter: i think you are correct !

code_generation/configuration.py
harrypuuter: :+1:

code_generation/configuration.py
harrypuuter: :+1:

code_generation/configuration.py
harrypuuter: nvm, i remember why this was in place here, we also delete the scope from self.scopes, which is also the list we use to iterate the scopes if we do not assign it a new list, thats why here, a new list is created. I will add a comment in the code to mention this.

nshadskiy (approved):

config/config.py
harrypuuter: Yes, i adapted the code, so that it is possible to write just `"global"` here, now it looks much more reasonable

harrypuuter:
duplicate of #114 --> Closed

Issue #115, opened 2021-11-25
[BUG] Jet pt ordering


nshadskiy:
**Describe the bug**
The algorithm to sort jets by their transverse momentum is working wrong.
https://github.com/KIT-CMS/CROWN/blob/468666c6cee15aef714b03460c3f3e9c0ed25c04/src/jets.hxx#L75-L81
`VecOps::Argsort` is sorting here the indices of the jets with non-zero pt and not the jet pt itself. This leads to a wrong ordering.

nshadskiy:
I made a pull request to fix this problem https://github.com/KIT-CMS/CROWN/pull/116

harrypuuter:
Resolved by #116

PR #116, opened 2021-11-25
fixing jet pt ordering

nshadskiy:
fixing jet pt ordering problem https://github.com/KIT-CMS/CROWN/issues/115
and adding new quantities


harrypuuter (approved):
very nice, looks good !

PR #120, opened 2021-12-22
Add Documentation for Configuration

harrypuuter:
Resolves #76

PR #123, opened 2022-01-12
add functionality to the shifts option

### LINKED ISSUE #122: [Request] cmake settings concerning shifts
Add some way of configuration to the cmake build command, to make it easy to 

1. produce nominal only
2. produce only a specific shift
3. produce only a set of shifts

harrypuuter:
Resolves #122 

The PR adds functionality to the `-DSHIFTS` option, which was only a dummy before. Now, the user can use this option to run only on a subset of systematic shifts or even no shift:
1.  `-DSHIFTS=none` runs nominal only, without any shifts
2.  `-DSHIFTS=all` runs all shifts only, this is the default if no option is provided
3.  `-DSHIFTS=comma,seperated,list,of,shifts` runs all shifts that match any of the strings in the list. So `-DSHIFTS=tauES` would match all shifts containing `tauES` in the name, including both the up and down variant. 

Additionally, some cleanup of the `generate.py`script was done, an the looping over multiple eras or samples during the code generation was moved to the cmake file, eliminating the python for loop in `generate.py`

As soon as #120 is merged, this update to the cmake command should also be documented there

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -15,7 +15,7 @@ env:
   SAMPLES: dy
   ERAS: 2018
   CHANNELS: "mt,mm"
-  SHIFTS: "tauES"
+  SHIFTS: "tauES,metUnclusteredEnUp"
 
 jobs:
   build_project:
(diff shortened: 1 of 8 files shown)


nshadskiy (approved):
Looks good!

PR #127, opened 2022-01-25
update triggerbits to the correct values and update according docs

### LINKED ISSUE #126: [BUG] Trigger filterbit documentation inconsistency
**Describe the bug**
When doing trigger object matching, a user can set a `filterbit` which should correspond to the filterbits mentioned in the docs: https://crown.readthedocs.io/en/latest/namespaces.html#_CPPv47trigger and moreover in the CMS code for the generation of the filterbits: https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/NanoAOD/python/triggerObjects_cff.py#L17

However, the values listed here are shifted by 0, and the value zero is also used for the first bit used, so the default value for no filterbit matching should be moved to `-1`

harrypuuter:
Resolves #126 

Update the docs to the correct values and set the default for no matching to `-1`, since 0 can be used as a triggerbit


harrypuuter:
New behaviour is now:

```
[2022-01-25 14:26:01.199] [CheckTriggerMatch] [warning] bit Check: true                                                                                                                       
[2022-01-25 14:26:01.199] [CheckTriggerMatch] [warning] bit cut: 4                  
[2022-01-25 14:26:01.199] [CheckTriggerMatch] [warning] bit Value: 00000000000000010000                                                                                                       
[2022-01-25 14:26:01.199] [CheckTriggerMatch] [warning] raw Value: 16  
```
where a check of triggerbit `4` should be true, if the raw value of 16 is set: https://github.com/cms-sw/cmssw/blob/ec200b2e6d0213cdf868308a7a836be8912c4cbc/PhysicsTools/NanoAOD/python/triggerObjects_cff.py#L72

nshadskiy (approved):
Looks good!

PR #132, opened 2022-02-09
Fix bug in trigger producer when running doubletriggers

harrypuuter:


ralfschmieder (approved):
Fix works. But only temporary, since in new nanoaod files the filterbits for all filters are available

PR #145, opened 2022-04-12
set default optimization to O2

harrypuuter:
A perfomance test showed, that switching from `-O3` to `-O2` has no real performance loss, but a much faster build time: 

build was full uncertainty model, et,mt,tt,em channel combined (~ 8000 Defines)

(runtime tested on a 1.3 Mio Event sample)
| Mode             | Build Time   | Runtime  |
| ------------- |-------------| ---------|
| `O3`              |  55399s (15:34 h) | 952 s |
| `O2 `             |    2564s (43 min) |928 s |
| `no optimization`|    136s            | 1940s |

Depending on the case, an optimization using `-O3` is still benefitial, but i suspect, that if the complexity becomes to big, the compiler optimization will not result in faster runtimes any longer, but only result in much much longer build times --> therefore set the default to `-O2` for a good middle ground.


nshadskiy (approved):
Seems reasonable!

PR #154, opened 2022-05-17
Jet energy corrections

nshadskiy:
The following thing are implemented with correctionlib using the newly added [json files](https://gitlab.cern.ch/cms-nanoAOD/jsonpog-integration/-/tree/master/POG/JME) provided by JetMET POG:
* [x] JER correction and uncertainty 
  * only AK4 jets for now, no JER for AK8 jets yet
* [x] Option for JES recorrection added. Some inconsistencies in jet pt's found for already applied and reapplied JES. Example:
  ```
  JE scale: corr. jet pt 44.15625 to raw jet pt 42.107986 to recorr. jet pt 44.14991
  JE scale: corr. jet pt 39.5 to raw jet pt 37.667725 to recorr. jet pt 39.49706
  JE scale: corr. jet pt 21.4375 to raw jet pt 18.08789 to recorr. jet pt 21.449396
  ```
  reasons for that might be float precision or newer version of corrections
* [x] JES uncertainties added: total, full set and reduced set
  * reduced set is self implemented and not from the JetMET POG json's (not included yet) -> use with caution!
* [x] reapplying of jet energy corrections to data possible -> implemented as variation since no information about specific era run in data samples
* [x] HEM issue implemented as an additional jes uncertainty as suggested [here](https://hypernews.cern.ch/HyperNews/CMS/get/JetMET/2000.html) by JetMET group
* [x] propagation to MET is done with the [propagateJetsToMet](https://github.com/KIT-CMS/CROWN/blob/7bddeb655623195561439514355a8f555196125c/src/met.cxx#L256) function, this is enough since type 1 corrections are already applied in nanoAOD


harrypuuter:
does not built with 

```
Traceback (most recent call last):
  File "/work/sbrommer/ntuple_prototype/CROWN/generate.py", line 79, in <module>
    generator.run(args)
  File "/work/sbrommer/ntuple_prototype/CROWN/analysis_configurations/tau/generate.py", line 65, in run
    config = config.build_config(
  File "/work/sbrommer/ntuple_prototype/CROWN/analysis_configurations/tau/config.py", line 1312, in build_config
    add_jetVariations(configuration, available_sample_types)
TypeError: add_jetVariations() missing 1 required positional argument: 'era'
CMake Error at CMakeLists.txt:274 (message):
  Code Generation Failed - Exiting !
```

I will still look at the code now :)

harrypuuter:
Tested building with: `cmake .. -DANALYSIS=tau -DCONFIG=config -DSAMPLES=data,dy -DERAS=2018 -DSCOPES=mm`

harrypuuter:
I think you also have to update the xpog correction submodule,

nshadskiy:
The submodule was updated in the first commit. Or do I have to do something else?

harrypuuter:
yes my bad, I forgot to update submodules on my checkout ..

src/jets.cxx
```diff
@@ -177,149 +179,315 @@ ROOT::RDF::RNode CutPUID(ROOT::RDF::RNode df, const std::string &maskname,
...
-                    if (deltaR > min_dR)
-                        continue;
-                    if (deltaR < 0.4 &&
-                        std::abs(pt_values_corrected.at(i) -
-                                 gen_pt_values.at(j)) >
-                            3.0 * reso * pt_values_corrected.at(i)) {
-                        min_dR = deltaR;
-                        genjetpt = gen_pt_values.at(j);
+                    ->debug("No gen jet found. Applying stochastic smearing.");
+                TRandom3 randm = TRandom3(
```
harrypuuter, line 336: Does this random number generator need to be initialized beforehand ?

src/jets.cxx
```diff
@@ -177,149 +179,315 @@ ROOT::RDF::RNode CutPUID(ROOT::RDF::RNode df, const std::string &maskname,
...
+                ROOT::Math::RhoEtaPhiVectorF genjet(gen_pt_values.at(j),
+                                                    gen_eta_values.at(j),
+                                                    gen_phi_values.at(j));
+                Logger::get("JetEnergyResolution")
+                    ->debug("Checking gen Jet:  Eta: {} Phi: {}", genjet.Eta(),
+                            genjet.Phi());
+                auto deltaR = ROOT::Math::VectorUtil::DeltaR(jet, genjet);
+                if (deltaR > min_dR)
+                    continue;
+                if (deltaR < (0.4 / 2.) &&
```
harrypuuter, line 317: what is this 0.4 /2. value here ? Is this always like this or only for AK4 ?

harrypuuter (commented):
very nice work ! I only had to minor understanding questions

src/jets.cxx
nshadskiy: That's true. This should depend on the algorithm. So 0.8 for AK8. I will change that.

harrypuuter (approved):

PR #155, opened 2022-05-23
restructure configurations to be plug and play

harrypuuter:
The idea is to move all python configurations out of the core repository into separate analysis repositories

This includes 

1. The configurations
2. Definition of Producers in Python
3. Definition of Quantities in Python

This way, the framework should be easier to maintain for multiple analyses at the same time. Shared code would be 

1. C++ functions
2. correction files 
3. Codebase for the automatic code generation 

while the rest would be independent of the main repository.

There are that many code changes, since the template analysis and the unittest analysis now need their own set of python producers 

In another PR, I will remove the configurations located in the `tau` folder completely and move them to another repository

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -11,7 +11,8 @@ on:
 # Settings go here!
 env:
   BUILD_TYPE: Release
-  ANALYSIS: test.unittest_config
+  ANALYSIS: unittest
+  CONFIG: unittest_config
   SAMPLES: dy
   ERAS: 2018
   SCOPES: "mt,mm,et"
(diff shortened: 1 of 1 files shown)


src/pairselection.cxx
```diff
@@ -838,7 +840,8 @@ ROOT::RDF::RNode PairSelection(ROOT::RDF::RNode df,
     Logger::get("tautau::PairSelection")
         ->debug("Setting up TauTau pair building");
     auto df1 = df.Define(
-        pairname, pairselection::fullhadronic::PairSelectionAlgo(mindeltaR),
+        pairname,
+        ditau_pairselection::fullhadronic::PairSelectionAlgo(mindeltaR),
```
ralfschmieder, line 844: shall we name this function fullyhadronic?

ralfschmieder (approved):

src/pairselection.cxx
harrypuuter: I would keep it as fullhadronic :)

PR #159, opened 2022-05-27
fix type in tauES variation definitions and add check 

harrypuuter:


nshadskiy (approved):
Looks good! But python formatting is failing.

PR #165, opened 2022-05-31
Add early Run 3 configurations as a submodule, add few more functions, bug fix in code_generation/configuration.py

khaosmos93:
- Configuration for Early Run 3: https://github.com/khaosmos93/CROWN-config-earlyRun3 → `analysis_configurations/earlyrun3`
- Few more simple functions:
   - `physicsobject::IsEmptyFlag`
   - `physicsobject::CutNFlag`
   - `physicsobject::SelectedObjects`
   - let me know if you are not happy with the naming / description
- minor bug fix in [code_generation/configuration.py](https://github.com/KIT-CMS/CROWN/blob/main/code_generation/configuration.py#L643-L652)


harrypuuter:
Looks good ! 

one thing i would change is to not add your config as a submodule to the CROWN repo, but include it only via a seperate clone similar to here: https://github.com/KIT-CMS/CROWN/blob/b554cea7c994baff72b425c4212ac439b6e520f6/init.sh#L31

This way you can update you config constantly and independent from the CROWN repo and we do not have to update the CROWN repo to always use the correct configuration commit

khaosmos93:
Thanks a lot for the comment. I updated it accordingly.

harrypuuter (approved):

PR #166, opened 2022-06-01
implementation of btagging scale factors with correctionlib

nshadskiy:
Implementing a function to calculate event weights to correct the shape of the b-tagging discriminator (DeepJet).

### DIFF
--- a/src/scalefactors.cxx
+++ b/src/scalefactors.cxx
@@ -647,6 +647,120 @@ ROOT::RDF::RNode id(ROOT::RDF::RNode df, const std::string &pt,
         {pt, eta});
     return df1;
 }
+
 } // namespace electron
+namespace jet {
+/**
+ * @brief Function used to evaluate b-tagging scale factors of jets with
+ * correctionlib, configurations:
+ * - [UL2018 b-tagging
+ * ID](https://cms-nanoaod-integration.web.cern.ch/commonJSONSFs/BTV_btagging_Run2_UL/BTV_btagging_2018_UL.html)
+ * - [UL2017 b-tagging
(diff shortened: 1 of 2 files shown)


src/scalefactors.cxx
```diff
@@ -647,6 +647,120 @@ ROOT::RDF::RNode id(ROOT::RDF::RNode df, const std::string &pt,
...
+            // bjets might have different cuts depending on the analysis
+            if ((jet_mask.at(i) || bjet_mask.at(i)) && jet_veto_mask.at(i)) {
+                Logger::get("btagSF")->debug(
+                    "SF - pt {}, eta {}, btag value {}, flavor {}",
+                    pt_values.at(i), eta_values.at(i), btag_values.at(i),
+                    flavors.at(i));
+                float jet_sf = 1.;
+                // considering only phase space where the scale factors are
+                // defined
+                if (pt_values.at(i) >= 20.0 && pt_values.at(i) < 10000.0 &&
```
harrypuuter, line 717: This is something to keep an eye out for future versions of the producer, since this might change with a different version of corrections. We could think about making these config parameters as well

src/scalefactors.cxx
```diff
@@ -647,6 +647,120 @@ ROOT::RDF::RNode id(ROOT::RDF::RNode df, const std::string &pt,
...
+                    std::abs(eta_values.at(i)) < 2.5) {
```
harrypuuter, line 718: same here

harrypuuter (commented):
only a minor suggestion to think about

src/scalefactors.cxx
nshadskiy: I think the scale factor phase space will not change anymore, at least for Run2. But sure we can change it to be defined in the config.

harrypuuter: i would say it is fine, but something to keep in mind :)

harrypuuter (approved):

PR #169, opened 2022-06-08
update trigger function to use regex matching and not crash, if a trigger is not found

harrypuuter:
In the current version of the trigger producer, the executable crashes, if a trigger path is not found in the input file. In data, it can happen, that a trigger path is only available in a fraction of all events (e.g. the `hps` triggers in 2018). 
With this PR, the trigger producer checks a given hltpath against all available hltpaths. 

	If no match is found, false is returned for all events
	If one match is found, the behavior is identical to before
	if more the one match is found and exception is thrown for now (to be implemented in the future)


khaosmos93 (approved):
Looks good!

PR #170, opened 2022-06-15
PU weights from JSON files

khaosmos93:
New producer for PU weights, the weight are directly from the LUM POG json files, e.g. `data/jsonpog-integration/POG/LUM/2018_UL/puWeights.json.gz`

PR tested with `earlyrun3` workflow.


harrypuuter (approved):
looks good 👍

PR #171, opened 2022-06-22
add check if inputfiles exists, add percentage to logging output

harrypuuter:
This PR adds a check, if all input files exists and can be opened, otherwise and error is thrown before running the event loop.

In addition, a sanity check to only add each output quantity once is added .

### DIFF
--- a/code_generation/analysis_template.cxx
+++ b/code_generation/analysis_template.cxx
@@ -48,9 +48,22 @@ int main(int argc, char *argv[]) {
         Logger::get("main")->info("Running with {} input files", argc - 2);
     }
     std::vector<std::string> input_files;
+    int nevents = 0;
     for (int i = 2; i < argc; i++) {
-        Logger::get("main")->info("input_file {}: {}", i - 1, argv[i]);
         input_files.push_back(std::string(argv[i]));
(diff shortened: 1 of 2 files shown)


code_generation/analysis_template.cxx
```diff
@@ -48,9 +48,22 @@ int main(int argc, char *argv[]) {
         Logger::get("main")->info("Running with {} input files", argc - 2);
     }
     std::vector<std::string> input_files;
+    int nevents = 0;
     for (int i = 2; i < argc; i++) {
```
nshadskiy, line 52: Why does it start with 2? Does the argv counter start with 1? I thought the the first input file is at position >=1 and the output file at position 0.
Also [here](https://github.com/KIT-CMS/CROWN/blob/77d02ce24e8af554c4665f8d470abbbac5388f84/code_generation/analysis_template.cxx#L47) shouldn't this be >=3 and not >3?

nshadskiy: Then never mind, thanks for clarifying. But the second comment with >=3 is still valid, right? You could have only one input file, then argc would be 3 and this case is not covered.

harrypuuter: I will just remove the printout

PR #177, opened 2022-07-19
fix jetID mask implementation

harrypuuter:
There is an error in the jetID mask application, to not include jets fulfilling a tighter ID

### DIFF
--- a/include/basefunctions.hxx
+++ b/include/basefunctions.hxx
@@ -376,7 +376,9 @@ inline auto FilterID(const int &index) {
 /// \returns a lambda function to be used in RDF Define
 inline auto FilterJetID(const int &index) {
     return [index](const ROOT::RVec<Int_t> &IDs) {
-        ROOT::RVec<int> mask = IDs == index;
+        ROOT::RVec<int> mask = IDs >= index;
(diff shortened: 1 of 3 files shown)


include/basefunctions.hxx
```diff
@@ -376,7 +376,9 @@ inline auto FilterID(const int &index) {
 /// \returns a lambda function to be used in RDF Define
 inline auto FilterJetID(const int &index) {
     return [index](const ROOT::RVec<Int_t> &IDs) {
-        ROOT::RVec<int> mask = IDs == index;
+        ROOT::RVec<int> mask = IDs >= index;
```
nshadskiy, line 379: I think the reason why I did this is that the [recommendation](https://twiki.cern.ch/twiki/bin/view/CMS/JetID13TeVUL#Preliminary_Recommendations_for) is to use the lepVeto and then you have only one option, so there shouldn't be any difference. But it might generally be better to do it like this.

nshadskiy: I see, I thought it was set to `6`. Then I mixed something up I guess and this should be fixed. I think to use `>=2` should also be alright for us since we clean the jets from the leptons ourselves.

PR #178, opened 2022-07-24
add extrapolation factor to trigger SF readout function 

harrypuuter:
This is used to apply tivial shifts of the trigger SF


ralfschmieder (approved):
looks great

PR #179, opened 2022-08-02
Add generic Producer to check dR overlap

harrypuuter:
This PR adds a Producer that can be used to generate a veto boolean, if a given p4 vector is overlapping with any particle within a collection.

One possible example is a FSR photon veto, where a p4 vector can be checked against all photons passing a selection

### DIFF
--- a/src/physicsobjects.cxx
+++ b/src/physicsobjects.cxx
@@ -256,6 +256,72 @@ ROOT::RDF::RNode SelectedObjects(ROOT::RDF::RNode df,
                      {inputmaskname});
 }
 
+/**
+ * @brief Function used to veto a particle, if it is overlapping within a given
+ * DeltaR value with another particle within a mask. For the particle to test,
+ * the lorentz vector is used, for the veto, the mask and all input quantities
+ * needed to calculate the lorentz vector are used. Returns true if the particle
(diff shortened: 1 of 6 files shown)


src/physicsobjects.cxx
```diff
@@ -256,6 +256,72 @@ ROOT::RDF::RNode SelectedObjects(ROOT::RDF::RNode df,
...
+ * @param dR_cut The maximum dR to be used for the veto
+ * @return ROOT::RDF::RNode a dataframe containing the new veto flag
+ */
+
+ROOT::RDF::RNode DeltaRParticleVeto(
+    ROOT::RDF::RNode df, const std::string &output_flag, const std::string &p4,
+    const std::string &particle_mask, const std::string &particle_pt,
+    const std::string &particle_eta, const std::string &particle_phi,
+    const std::string &particle_mass, const float dR_cut) {
+    auto veto_overlapping_photon =
```
nshadskiy, line 289: Since this function is not only for photons I would rename this to `veto_overlapping` or maybe `veto_overlapping_particle`.

harrypuuter: done !

nshadskiy (approved):
Looks good!

PR #181, opened 2022-08-12
Triggerbit only matching / Embedding Updates

harrypuuter:
This PR adds a producer to create a boolean based on triggerobject matching only, so no passing hlt path. This is needed for tau embedded samples. 

Additionally, the embedding lepton corrections were updated. 

Additionally, TauID Scalefactors for embedding were added plus the corresponding producers


ralfschmieder (approved):
top

PR #186, opened 2022-08-26
allow the use of modifiers within Shift configs

harrypuuter:
This PR adds the possibility to use Modifiers within the configuration of shifts e.g.:

```py 
configuration.add_shift(
            SystematicShift(
                name="eleEsBarrelDown",
                shift_config={
                    ("global"): {
                        "ele_energyscale_barrel": EraModifier(
                            {"2018": ele_energyscale_2018["barrel"]["down"]}
                        )
                    }
                },
                producers={("global"): electrons.ElectronPtCorrectionEmbedding},
            ),
            samples=["embedding"],
        )
```
such that, within a systematic shift, different values can be used depending on the era / sample. Before, this modifiers would not be resolved, resulting in an error during the code generation

### DIFF
--- a/code_generation/configuration.py
+++ b/code_generation/configuration.py
@@ -338,6 +338,8 @@ class Configuration(object):
                 scopes_to_shift = [
                     scope for scope in shift.get_scopes() if scope in self.scopes
                 ]
+                # if a modifier is used within the shift config, we have to resolve it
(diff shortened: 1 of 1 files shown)


khaosmos93 (commented):
Looks good!

PR #188, opened 2023-01-09
add check for idential names, convert name to unique naming

### LINKED ISSUE #182: [BUG] Throw expection, if two producers have the same name, resulting in overwriting of generated code
If two producers have the same name, and live in the same scope, the code generation will overwrite the files generated from these producers, without any check. It should result in an error transparent to the user

harrypuuter:
Fixes #182 

This fix does not throw an exception but makes the generated names unique. If two code subsets with the same name are generated, the second one will be made unique before generating the code to avoid linking errors: 

```c++
auto df3_mt = GoodMuons_mt(df2_mt); 
[...]
auto df19_mt = GoodMuons_1_mt(df18_mt);
 ```

in addition, a warning will be printed for the user

### DIFF
--- a/code_generation/code_generation.py
+++ b/code_generation/code_generation.py
@@ -25,6 +25,7 @@ class CodeSubset(object):
         scope: The scope of the code generation.
         folder: The folder in which the code will be generated.
         parameters: The parameters to be used for the generation.
+        name: The name of the code subset.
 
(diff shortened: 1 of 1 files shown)


mburkart (approved):
The changes look sound to me.

PR #203, opened 2023-03-15
additional, more generic function for b tagging scale factors

nfaltermann:
extended initial functionality to a more generic use case, less analysis-dependent

also some minor refactoring and typo fixing in the already existing function


harrypuuter (approved):
I guess you know what you need :D

PR #206, opened 2023-03-27
trigger prescales

nfaltermann:
function for trigger prescale values from a JSON


harrypuuter (approved):
looks good and positive !

PR #212, opened 2023-04-17
add check that crown is run in the correct folder // Tautrigger

harrypuuter:
This PR adds check that the data folder is available, adds a Tautrigger producer and a fix for the ReplaceProducer function


ralfschmieder (approved):
great work

PR #213, opened 2023-04-17
add check for handeling of input files with zero events

harrypuuter:
This PR adds a modification to the code generation, to be able to handle input files with zero events in the ntuple. This is relevant for the automatic production of friend trees, where ntuple jobs can end up generating events with 0 events in them. 

In such a case, an empty ntuple file will be generated.


nshadskiy (approved):
Looks fine to me!

PR #219, opened 2023-04-28
add SM fakefactor functions, improve performance of nmssm functions

harrypuuter:


src/fakefactors.cxx
```diff
@@ -600,5 +600,472 @@ fakefactor_sm_lt(ROOT::RDF::RNode df, const std::string &outputname,
...
+    auto ttbar =
+        correction::CorrectionSet::from_file(ff_file)->at("ttbar_fake_factors");
+    auto fractions =
+        correction::CorrectionSet::from_file(ff_file)->at("process_fractions");
+
+    auto qcd_lep_pt_closure = correction::CorrectionSet::from_file(ff_corr_file)
+                                  ->at("QCD_non_closure_lep_pt_correction");
+    auto qcd_lep_iso_closure =
+        correction::CorrectionSet::from_file(ff_corr_file)
+            ->at("QCD_non_closure_lep_iso_correction");
```
nshadskiy, line 752: Since I renamed some of the corrections in the FF framework (latest commit) this might not be correct anymore. (this also applies for other corrections).

harrypuuter: Can you point me to the naming changes ? :)

nshadskiy: I think you can look at this dict here https://github.com/nshadskiy/TauFakeFactors/blob/main/helper/correctionlib_json.py#L8-L25
I basically added `leading` or `subleading` to the naming corresponding to first and second lepton (e, mu, tau). I think I did it like this because it was easier to be consistent in the tt channel.

nshadskiy: To be save you can look at the summary of the correctionlib files you are using. It just has to be consistent with the naming there.

nshadskiy:
PR closed because outdated

PR #220, opened 2023-05-16
fixed small bug: double semicolon

ralfschmieder:
bug in src/lorentzvectors: there were two semicolons instead of one


nshadskiy (approved):
Nice catch :D

PR #223, opened 2023-06-21
implemetation of the ditau trigger SF producer for embedded events

nshadskiy:
The added json file is only preliminary. There will be an official file after the validation of the scale factors.


data/embedding/tau_trigger2018_UL.json.gz
harrypuuter, line 1: Is this file provided by the tauPOG or did you convert it by yourself ?

PR #210, opened 2023-04-04
add SM fakefactor functions, improve performance of nmssm functions

harrypuuter:

PR #224, opened 2023-06-21
implementation of fake factors for full hadronic channel

nshadskiy:


harrypuuter (changes requested):
looks fine by me, one small remark, going forward I would like to switch to the @param instead of \param for the documentation, since this is handled a bit better by doxygen and clang-format

harrypuuter:
do you already use the changes in #210 concerning the performance of the semileptonic channels ?

nshadskiy:
I changed it now to @param. And yes, https://github.com/KIT-CMS/CROWN/pull/210 is actually already included.

harrypuuter (approved):

PR #229, opened 2023-08-25
added tau ID 2018 UL scale factors

conformist89:
added tau ID 2018 UL scale factors


harrypuuter:
do you also have a correctionlib file that contains all measured wps ? I guess this would be beneficial for other analyses like @nshadskiy or @ralfschmieder

conformist89:
> do you also have a correctionlib file that contains all measured wps ? I guess this would be beneficial for other analyses like @nshadskiy or @ralfschmieder

No, I don't have the whole set of wps. This file includes only  all the vs Jet wps with Tight and VVLoose Vs Ele but not  for all the vsEle wps

harrypuuter:
Ah i see, sorry i was confused by the `Tight` in the filename and thought this would only contain the `Tight` vsJet WP

PR #230, opened 2023-08-28
add functionality to use arbitrary number of friend trees

harrypuuter:
Before, only a CROWN ntuple could be used as input for friend trees. With this PR, a separate template is used for friend trees executables, which allows to use an arbitrary amount of friend files to be used. This is important e.g. for the generation of NN outputs 

```
./executable outputfile.root crown_ntuple.root friend_1.root friend_2.root friend_3.root
```

During the cmake build step, the quantitiesmaps for each file have to be provided as a comma-separated list using the `-DQUANTITIESMAP` (providing either a ROOT file or a json file)

```
cmake .. -DANALYSIS=unittest -DCONFIG=unittest_friends_2 -DSAMPLES=dyjets -DERAS=2018 -DSCOPES=mt -DSHIFTS=all -DQUANTITIESMAPS=/path/to/crown_ntuple.root,/path/to/friend_1.root,/path/to/friend_2.root,/path/to/friend_3.root  -DDEBUG=false

```

In addition, this PR fixes some issues with friend trees:

- Incorrect treatment of shift propagation, now shifts are appropriately set and only required shifts are executed
- Validation of input files. How a scan of all required inputs is performed, catching potential misconfiguration before compilation.


harrypuuter:
TODO: add unittest for this case

ralfschmieder (approved):
works fine!

PR #234, opened 2023-11-02
Addition of updated correctionlibs for 2016post/preVFPUL

a-monsch:
As the title lets you guess.


harrypuuter (approved):

harrypuuter:
Ah one this note, maybe we change this now, to split the corrections into two files, one for muons and one for the embedding specific corrections ?

a-monsch:
Done a split of muon correctionlib into muon and embeddingselection.

PR #235, opened 2023-11-11
ML inference

nfaltermann:
added functionality to evaluate ML models from Keras and PyTorch via CROWN, utilizing JIT compilation from SOFIE interface

WIP


nfaltermann:
currently not working with shifts, looking into it

nfaltermann:
more general implementation now available via ONNX Runtime, therefore close

PR #242, opened 2024-01-19
add exeption for misconfiguration of sampleypes

harrypuuter:
Now, CROWN will throw an error during configuration, when a rule is configured to be applied to a nonexistent sampletype:

```
code_generation.exceptions.SampleRuleConfigurationError: Sampletype embedding,embedding_mc cannot be used in Rule ProducerRule - add [ProducerGroup: RenameJetsData] for ['embedding,embedding_mc'] in scopes ['global'] since the type is not defined. Available samples types are {'ttbar', 'embedding', 'embedding_mc', 'wjets', 'singletop', 'data', 'ggh_hbb', 'vbf_hbb', 'rem_hbb', 'dyjets', 'diboson', 'rem_htautau', 'ggh_htautau', 'vbf_htautau', 'electroweak_boson'}
CMake Error at CMakeLists.txt:338 (message):
  Code Generation Failed - Exiting !
  ```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -45,7 +45,7 @@ jobs:
       - name: install packages
         run: apt-get install -y git cmake make python3-pip openmpi-bin libopenmpi-dev libboost-all-dev openssh-client
       - name: install python packages
-        run: python3 -m pip install GitPython --break-system-packages && python3 -m pip install git+https://github.com/cms-nanoAOD/correctionlib.git --break-system-packages
+        run: python3 -m pip install GitPython && python3 -m pip install git+https://github.com/cms-nanoAOD/correctionlib.git
 
       - name: Clone project
         uses: actions/checkout@v3
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -43,7 +43,7 @@ jobs:
       - name: update ubuntu
         run: apt-get -y update
       - name: install packages
-        run: apt-get install -y git cmake make python3-pip openmpi-bin libopenmpi-dev libboost-all-dev openssh-client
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by harrypuuter** [comment]
This should catch some errors, ofc things like 

```
samples=[
            sample
            for sample in available_sample_types
            if sample not in ["data, embedding", "embedding_mc"]
        ],
```
are still not checked by this

**Comment by a-monsch** [comment]
> This should catch some errors, ofc things like
> 
> ```python
> samples=[
>             sample
>             for sample in available_sample_types
>             if sample not in ["data, embedding", "embedding_mc"]
>         ],
> ```
> 
> are still not checked by this

Would it make sense to add somewhere a check like searching for `'(".\w+.,.\w+.")'` to catch those obvious but from time to time hardly detectable mistakes?

**Comment by harrypuuter** [comment]
I don't want to add such checks to the framework, since regex checking a configuration is not something a framework should do. On the other hand, the case that is now checked is something well within the information given to CROWN by the configuration, this second way of setting up rules is not really intended. 

What would make more sense would be to extend the capabilities of ProducerRules so the user can set the samples, where the rule should not be applied, something like 

```py
configuration.add_shift(
        SystematicShiftByQuantity(
            name="metUnclusteredEnUp",
            quantity_change={
                nanoAOD.MET_pt: "PuppiMET_ptUnclusteredUp",
                nanoAOD.MET_phi: "PuppiMET_phiUnclusteredUp",
            },
            scopes=["global"],
        ),
        exclude_samples= ["data", "embedding", "embedding_mc"],
    )
```

which would then mean, we can validate the list again, and the Rule is automatically applied to all samples that are not excluded

a-monsch (approved):

PR #248, opened 2024-02-07
readout of the emb electron energy scale with correctionlibs

ralfschmieder:
correctionlibs for Run-2UL emb electron energy scale
made a new function, to read out the sfs for the emb electron energy scales
remark: 2017 is not ready yet and only filled with dummy values


src/physicsobjects.cxx
```diff
@@ -968,6 +968,55 @@ PtCorrection_byValue(ROOT::RDF::RNode df, const std::string &corrected_pt,
         df.Define(corrected_pt, electron_pt_correction_lambda, {pt, eta});
     return df1;
 }
+/// Function to correct electron pt, based on correctionlib file
+///
+/// \param[in] df the input dataframe
+/// \param[out] corrected_pt name of the corrected tau pt to be calculated
+/// \param[in] pt name of the raw tau pt
+/// \param[in] eta name of raw tau eta
+/// \param[in] sf_barrel scale factor to be applied to electrons in the barrel
```
harrypuuter, line 977: thats the name of the sf right? so something like `sf_barrel_name` would maybe be better

src/physicsobjects.cxx
```diff
@@ -968,6 +968,55 @@ PtCorrection_byValue(ROOT::RDF::RNode df, const std::string &corrected_pt,
...
+/// \param[in] sf_endcap scale factor to be applied to electrons in the endcap
```
harrypuuter, line 978: same here

harrypuuter (approved):
just two naming remarks

PR #252, opened 2024-02-29
restructure cmake files into multiple files

harrypuuter:
The PR restructures the cmake file into multiple files in the cmake folder, since the `CMakeLists.txt` got rather long and hard to read over the last months


harrypuuter:
Also, onnxruntime is added as a lib to be used in CROWN

ralfschmieder (approved):
works fine

PR #254, opened 2024-03-07
Add working Onnxruntime Producer

harrypuuter:
This PR adds a generic implementation of an OnnxRuntimeProducer


code_generation/code_generation.py
```diff
@@ -106,8 +106,9 @@ def write(self):
         log.debug("Writing code subset {}".format(self.name))
         log.debug("folder: {}, file_name: {}".format(self.folder, self.file_name))
         # write the header file if it does not exist or is different
-        with open(self.headerfile + ".new", "w") as f:
-            f.write(f"ROOT::RDF::RNode {self.name}(ROOT::RDF::RNode df);")
+        f.write(
```
nshadskiy, line 109: As Codacy is saying, `f` is not defined anymore. You still need this line before, right?
```python
with open(self.headerfile + ".new", "w") as f:
```

code_generation/analysis_template.cxx
```diff
@@ -104,6 +106,9 @@ int main(int argc, char *argv[]) {
     // file logging
     Logger::enableFileLogging("logs/main.txt");
 
+    // start an onnx session manager
+    OnnxSessionManager onnxSessionManager;
```
nshadskiy, line 110: Why don't we need this in the friend template?

harrypuuter: fixed

include/ml.hxx
```diff
@@ -0,0 +1,96 @@
...
+
+    if (nParameters != InputList.size()) {
+        Logger::get("OnnxEvaluate")
+            ->error("Number of input parameters does not match the number of "
+                    "input variables: {} vs {}",
+                    nParameters, InputList.size());
+        throw std::runtime_error("Number of input parameters does not match");
+    }
+
+    std::string replace_str = std::string("EVTID");
```
nshadskiy, line 56: Maybe add a comment here that this is only needed if an even/odd splitting used.

harrypuuter: Your right, in the generic version, this is not required

include/ml.hxx
```diff
@@ -0,0 +1,96 @@
...
+        std::vector<float> output = onnxhelper::run_interference(
+            session, allocator, inputs, input_node_dims, output_node_dims,
+            num_input_nodes, num_output_nodes);
+
+        timer.Stop();
+        Logger::get("OnnxEvaluate")
+            ->debug("Inference time: {} mus", timer.RealTime() * 1000 * 1000);
+        return output;
+    };
+    auto df3 = df2.Define(outputname,
```
nshadskiy, line 89: I don't see any definition of `df2`, is that correct?

harrypuuter: It should be 

```
df1 = df.Define(outputname, 
```

include/utility/OnnxSessionManager.hxx
```diff
@@ -0,0 +1,58 @@
...
+    };
+    Ort::Session *getSession(const std::string &modelPath) {
+        // check if session already exists in the sessions map
+        if (sessions_map.count(modelPath) == 0) {
+            sessions_map[modelPath] = std::make_unique<Ort::Session>(
+                env, modelPath.c_str(), session_options);
+            Logger::get("OnnxSessionManager")
+                ->info("Created session for model: {}", modelPath);
+            // std::unique_ptr<Ort::Session> session_ptr(session);
+            // sessions_map[modelPath] = std::move(session_ptr);
```
nshadskiy, line 28: Commented line can be dropped, I guess.

src/ml.cxx
```diff
@@ -0,0 +1,150 @@
...
+ * @param df The input dataframe
+ * @param inputname name of the variable which should be transformed
+ * @param outputname name of the output column
+ * @param param_file path to a json file with a dictionary of mean and std
+ * values
+ * @param var_type variable data type for correct processing e.g. "i" for
+ * integer or "f" for float
+ * @return a new dataframe containing the new column
+ */
+ROOT::RDF::RNode StandardTransformer(ROOT::RDF::RNode df,
```
nshadskiy, line 49: This producer is also trimmed for even/odd evaluation and I also changed the format of the input information file. While I used json files, in https://github.com/KIT-CMS/CROWN/pull/235 csv files are used. So I think either a common definition should be defined for CROWN or the producer needs to be expanded to be more generic and able to take different input formats.

harrypuuter: We could also rename it to json in the name and the other implementation to csv in the name to have multiple functions available. I am not a fan of depending on something like file endings

src/utility/OnnxSessionManager.cxx
```diff
@@ -0,0 +1,122 @@
...
+    Ort::MemoryInfo memory_info = Ort::MemoryInfo::CreateCpu(
+        OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault);
+
+    std::vector<Ort::Value> inputTensors;
+    std::vector<Ort::Value> outputTensors;
+
+    size_t inputTensorSize = vectorProduct(input_node_dims);
+    size_t outputTensorSize = vectorProduct(output_node_dims);
+
+    // outputTensorSize = 8;
```
nshadskiy, line 31: Can be dropped.

PR #258, opened 2024-04-11
add error handeling for validation errors

harrypuuter:


a-monsch:
@harrypuuter The two build checks are dying (at different stages...). How critical are they?

harrypuuter:
This is a problem with the ROOT docker images, I will try to fix this in another PR

PR #269, opened 2024-06-05
fix FastMTT memory leak

harrypuuter:
This PR fixes the FastMTT memory leak observed in the current main function

Before:
grafik

After:
grafik


nshadskiy (approved):
Looks fine to me! See one small comment.

PR #270, opened 2024-06-07
update top pT reweighting

nfaltermann:
updated to the latest recommendation from:
 https://twiki.cern.ch/twiki/bin/view/CMS/TopPtReweighting#TOP_PAG_corrections_based_on_dat


harrypuuter (approved):
thanks!

PR #271, opened 2024-07-05
removal of wrong pileup weights function

nfaltermann:
the version of the function that takes ROOT files as input produces incorrect results.

since we moved to correctionlib there is no need for this to be fixed, thus we can delete it. at least it should not be used in the template analysis anymore.


harrypuuter (approved):
looks good 👍🏼

Issue #264, opened 2024-05-06
Drop CentOS8 Support


harrypuuter:
Once ETP has migrated to Alma9, we can start to drop CS8 support (finally)

harrypuuter:
In Kingmaker, Support will be dropped with this PR: https://github.com/KIT-CMS/kingmaker-images/pull/6 - drop support in CROWN as well as soon as merged

harrypuuter:
done with #275

PR #275, opened 2024-07-23
switch to lcg 106

harrypuuter:
This PR updates some CROWN dependencies

- [x] Switch to LCG 106 and GCC 13
- [ ] Switch to C++ 20 Standard
- [x] Switch to spdlog 1.14.1
- [x] Drop RHEL 8 Support


harrypuuter:
fixes #264

include/utility/Logger.hxx
```diff
@@ -1,12 +1,96 @@
...
+// Register the formatter with fmt library for any size of std::bitset
+template <std::size_t N>
+struct fmt::formatter<std::bitset<N>> : formatter<std::string> {
+    template <typename FormatContext>
+    auto format(const std::bitset<N> &b,
+                FormatContext &ctx) -> decltype(ctx.out()) {
+        return BitsetFormatter<N>::format(ctx, b);
+    }
+};
+
```
nshadskiy, line 93: Does this already cover all the cases that we have in Crown?

harrypuuter: most basic types are covered, only the ones that are not "common" like the ROOT special types. At least the tau analysis does compile with this. Maybe other analyses need additional types here

nshadskiy (approved):
Looks good to me!

Issue #279, opened 2024-07-26
ReplaceProducer doesn't add outputs if outputs of old and new producer are different


moritzmolch:
I want to replace a ProducerGroup with ReplaceProducer for a scope (in this example the `mm` scope). However, when I do that the outputs of the new producer are not showing up in the list of all outputs (the list of branch names in the output file of the respective scope.

The old producer is
```python
BasicBJetQuantities = ProducerGroup(
    name="BasicBJetQuantities",
    call=None,
    input=None,
    output=None,
    scopes=["mt", "et", "tt", "em", "mm", "ee"],
    subproducers=[
        NumberOfBJets,
        NumberOfBJets_boosted,
    ],
)  # outputs nbtag, nbtag_boosted
```
with outputs `nbtag` and `nbtag_boosted` and the new producer is
```python
BasicBJetQuantitiesMuMu = ProducerGroup(
    name="BasicBJetQuantitiesMuMu",
    call=None,
    input=None,
    output=None,
    scopes=["mm"],
    subproducers=[
        NumberOfBJets,
    ],
)  # outputs nbtag
```
with the output `nbtag`. The modification in the main config file is declared with
```python
configuration.add_modification_rule(
    ["mm"],
    ReplaceProducer(
        producers=[jets.BasicBJetQuantities, jets.BasicBJetQuantitiesMuMu],
        samples=available_sample_types,
    ),
)
```

The problem is now that I cannot find `nbtag` as branch in the output file of the `mm` scope, although it is an output of the new producer. I also declared it with the `configuration.add_outputs` explicitly as output variable in the `mm` scope.

Also, the debug output of the `cmake` compilation of the configuration shows that the output `nbtag` is removed. The variable `nbtag_boosted` is not appearing here as I didn't declare it as an output of the `mm` scope in the main configuration, so I guess this part is okay. However, there is no line saying that `nbtag` is added back: 

```
...
[DEBUG   ] [rules.py           :303] scopes: ['mm']
[DEBUG   ] [rules.py           :304] Producers to be replaced: [ProducerGroup: BasicBJetQuantities, ProducerGroup: BasicBJetQuantitiesMuMu]
[DEBUG   ] [rules.py           :309] ReplaceProducer: Replace ProducerGroup: BasicBJetQuantities from producer in scope mm with ProducerGroup: BasicBJetQuantitiesMuMu
[DEBUG   ] [rules.py           :387] ReplaceProducer: Removing nbtag from outputs in scope mm
[...]

PR #280, opened 2024-07-26
Fix: ReplaceProducer doesn't add outputs if outputs of old and new producer are different

moritzmolch:
This pull request should fix https://github.com/KIT-CMS/CROWN/issues/279.


nshadskiy:
Closed because issue solved by https://github.com/KIT-CMS/CROWN/pull/303

Issue #268, opened 2024-05-29
[BUG] No "ntuple" tree in ntuples if 0 events selected


nshadskiy:
It might happen that no events are selected during ntupling. e.g. here for a data nanoAOD
```
[2024-05-25 13:25:33.386]   Starting Evaluation                                                                   CROWNRun.py:187
[13:25:52] [2024-05-25 13:25:52.076]   10000 / 56973 (17.55 %) Events processed ...                                          CROWNRun.py:187
         [2024-05-25 13:25:52.123]   et:                                                                                   CROWNRun.py:187
           GoldenJSONFilter: pass=0          all=56973      -- eff=0.00 % cumulative eff=0.00 %                              CROWNRun.py:187
           Flag_goodVertices: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                             CROWNRun.py:187
           Flag_globalSuperTightHalo2016Filter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %           CROWNRun.py:187
           Flag_HBHENoiseFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                          CROWNRun.py:187
           Flag_HBHENoiseIsoFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                       CROWNRun.py:187
           Flag_EcalDeadCellTriggerPrimitiveFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %       CROWNRun.py:187
           Flag_BadPFMuonFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                          CROWNRun.py:187
           Flag_eeBadScFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                            CROWNRun.py:187
           Flag_ecalBadCalibFilter: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                       CROWNRun.py:187
           GoodElTauPairs: pass=0          all=0          -- eff=-nan % cumulative eff=0.00 %                                CROWNRun.py:187
```
The output ntuple is still saved with all the information but since no events are selected no "ntuple" tree is saved. e.g.
```bash
root [1] .ls
[...]

PR #281, opened 2024-08-21
add additional check if ntuple tree is empty

nshadskiy:
This PR is addressing https://github.com/KIT-CMS/CROWN/issues/268

If a ntuple is processed but has no events left after all applied filters, the ntuple file does not contain a `ntuple` tree because it seems that the `rdf.Snapshot()` function does not write out  a tree if it's empty. This leads to the problem that friend production fails for these ntuples although they were processed correctly. To account for this an additional check is added to look at the `quantities` tree. This tree is saved even if the event number is 0 since it is not done by `rdf.Snapshot()`.


harrypuuter (approved):

nshadskiy:
resolves https://github.com/KIT-CMS/CROWN/issues/268

PR #282, opened 2024-10-02
progressbar only if nevents!=0

ralfschmieder:
progressbar results in error, if nevents=0 and no ntuple tree is available. Therefore the progressbar is only initiated if nevents!=0


nshadskiy (approved):
Nice catch!

PR #286, opened 2024-12-06
Merge some changes of CROWN_tutorial branch with main

tvoigtlaender:
- Replace "is not None" checks as lists/sets/dicts are not seen as "None", even when empty.

- Allow for renaming of global scope and usage of no global scope by changing the setting in the analysis config.

- Add empty build dir.


.gitignore
```diff
@@ -1,8 +1,9 @@
 # folders
-build/
-build*/
+build/*
+!build/.gitkeep
```
nshadskiy, line 3: Does this still cover the `build*/` case? I'm not that familiar with the gitignore syntax.

code_generation/configuration.py
```diff
@@ -88,7 +89,7 @@ def __init__(
         self.available_scopes = set(available_scopes)
         self.available_outputs: Dict[str, QuantitiesStore] = {}
         self.available_shifts: Dict[str, Set[str]] = {}
-        self.global_scope = "global"
+        self.global_scope = global_scope
```
nshadskiy, line 92: If you want to do this you probably also need propagate this everywhere where `"global"` is explicitly used. Like here
https://github.com/KIT-CMS/CROWN/blob/95ea5cb8f6bf33ed5dbfc7ab635587a093527a40/code_generation/producer.py#L149
https://github.com/KIT-CMS/CROWN/blob/95ea5cb8f6bf33ed5dbfc7ab635587a093527a40/code_generation/rules.py#L23
https://github.com/KIT-CMS/CROWN/blob/95ea5cb8f6bf33ed5dbfc7ab635587a093527a40/code_generation/optimizer.py#L82
as just a few examples.

.gitignore
tvoigtlaender: I think it doesn't, but I'm also unsure if build*/ dirs are ever exist. I thought all of the generated code is put into the build dir.

Do you have an example where build*/ is generated?

tvoigtlaender:
Some of the 'if cases' in the code make assumptions that lead to unintended consequences.

One such example: 
A check of whether a variable is 'None'. 
The actual value is an empty list and is therefore not 'None'. 
In most cases, the handling for an empty list and a 'None' would be the same.
However, a few select [places](https://github.com/KIT-CMS/CROWN/blob/95ea5cb8f6bf33ed5dbfc7ab635587a093527a40/code_generation/producer.py#L682) explicitly check for non-'None' empty values (like an empty list) later on and have to be treated accordingly.

.gitignore
nshadskiy: I often have multiple `build_*` directories when I test stuff with different configurations, e.g. a build folder for ntuples and a build folder for friends which take the produced ntuples. Or for comparisons of different ntuple configurations.

tvoigtlaender: Added 'build*/' back.

code_generation/configuration.py
tvoigtlaender: The intended change here is to only provide the option to not have a `global` scope at all. A more direct choice for this was implemented in [Link](https://github.com/KIT-CMS/CROWN/pull/286/commits/542ca2cdd09a8dc7cbc11bb865fd47daaaf72752#diff-bac2f5f7803facde8e079f690ad030af05a725fed5d366f97a2f0d41516b2430). As of now, there is no documentation for this feature.

Renaming the global scope is not really needed in my opinion.
It is currently not possible to only have a global scope due to a separate bug that is out of the scope of this PR.
The two cases lead to a very similar outcome (Only having a single scope), but the option to not have a global scope at all leaves more options for the naming of scopes. It also works with the case in which all scopes are treated differently from the very start and no global scope would be used.

code_generation/configuration.py
```diff
@@ -613,7 +618,7 @@ def _remove_empty_configkeys(self, config) -> None:
             Returns:
                 None
         """
-        for key in config:
+        for key in list(config):
```
nshadskiy, line 621: Why do we need to change it to a list?

build/.gitkeep
nshadskiy, line 1: This `build` folder should not be present since we ignore it, right?

nshadskiy (approved):
Besides the two understanding questions, I think everything is fine.

code_generation/configuration.py
nshadskiy: Ok, we should work on this is a separate PR. I will make an issue for that.

build/.gitkeep
tvoigtlaender: This `!build/.gitkeep` is excluded from the ignore. The idea is that you'd already have the build dir when you check out the repo, and you don't have to make it before you can start. Everything except the hidden file in `build` is still ignored.

code_generation/configuration.py
tvoigtlaender: Removed in https://github.com/KIT-CMS/CROWN/pull/286/commits/c6acbae4d9494be64640dcec1c85f52fd76d5a4f

code_generation/configuration.py
tvoigtlaender: Sounds reasonable. Removed in https://github.com/KIT-CMS/CROWN/pull/286/commits/6bd7f6372cd42c387c37124ef05cf8b053cf2408

PR #291, opened 2024-12-15
Fix for cpp_addons PR. Templating for subset_template fixed.

tvoigtlaender:
Templating is currently not applied to subset template.
This PR adds this functionality.
Also turns logging of include templating to debug level.


a-monsch:
Thank you!

PR #303, opened 2025-02-27
Adjusting functionality of ReplaceProducer to correctly handle outputs of ProducerGroups

nshadskiy:
This is a more correct fix to the issue https://github.com/KIT-CMS/CROWN/issues/279
And it is accounting for outputs of ProducerGroups that are or aren't written out to the ntuple when they are replaced, which is not fully covered by https://github.com/KIT-CMS/CROWN/pull/280


moritzmolch (approved):
Looks fine to me.

Issue #308, opened 2025-03-17
[Request] Add default values for the KingMaker workflow options


nshadskiy:
For the law run examples in the documentation the default values of some of the options are missing.
https://crown.readthedocs.io/en/latest/kingmaker.html

nshadskiy:
Resolved in PR https://github.com/KIT-CMS/CROWN/pull/310

PR #310, opened 2025-03-20
Add kingmaker default to documentation

tvoigtlaender:
Relating to https://github.com/KIT-CMS/CROWN/issues/308

I believe this is the only setting that was missing its default, as far as parameters that change frequently during the use of KingMaker goes.

Other settings are covered in the `KingMaker_luigi.cfg` file, but are currently not further explained in the [documentation](https://crown.readthedocs.io/en/latest/kingmaker.html#kingmaker-configuration).


nshadskiy (approved):
Nice, thanks!

PR #320, opened 2025-06-03
Dev pr bugfixes

ysfseday:
type of genmatch modified

### DIFF
--- a/include/jets.hxx
+++ b/include/jets.hxx
@@ -18,8 +18,9 @@ PtCorrectionMC(ROOT::RDF::RNode df,
                const std::string &jer_tag, bool reapply_jes,
                const int &jes_shift, const std::string &jer_shift);
 ROOT::RDF::RNode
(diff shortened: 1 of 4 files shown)


nshadskiy:
The build step is failing due to the change of the gen_match type.
```bash
2: terminate called after throwing an instance of 'std::runtime_error'
2:   what():  RDataFrame: type mismatch: column "tau_gen_match_2" is being used as int but the Define or Vary node advertises it as unsigned char
2/2 Test #2: crown_ntuple_unittest_config_dyjets_2018 ...Subprocess aborted***Exception:  21.30 sec
```
You also need to adjust the python producers in `analysis_configurations/unittest` and  `analysis_configurations/template_analysis`.

nshadskiy (approved):
Looks good. I suggest that we still merge it despite the error. I will work on only having one gen match definition in https://github.com/KIT-CMS/CROWN/pull/319

PR #324, opened 2025-07-06
Reintroducing tau gen matching with only 6 states

a-monsch:
Fix accompanied with https://github.com/KIT-CMS/TauAnalysis-CROWN/pull/63


nshadskiy (approved):
Looks good from my side.

PR #328, opened 2025-07-21
Functions for tau corrections in Run 3

moritzmolch:
Compared to Run 2, the tau energy corrections in Run 3 additionally depend on the DeepTau working points for ID vs electrons and ID vs jets. The scale factor for DeepTau ID vs electrons additionally depends on the decay mode of the tau. The functions in CROWN are adapted to this change:

- `PtCorrectionMC` is a new function for evaluating the tau energy corrections in Run 3. All steps, previously split in corrections for electron fakes, muon fakes, and genuine taus, are performed inside this single function now. The helper function `get_tes_variation` is used to return the correct variation, depending on the generator match to the reconstructed tau, the decay mode, and the region in the detector.
- An overloaded version `Id_vsEle` has been added, now also containing the decay mode of the tau as input parameter.

### DIFF
--- a/src/taus.cxx
+++ b/src/taus.cxx
@@ -224,29 +224,21 @@ PtCorrectionMC(
     auto evaluator =
         correction_manager.loadCorrection(es_file, correction_name);
 
-    // set the variation depending on the gen match, decay mode, and barrel/endcap region
-    std::string variation = get_tes_variation(
-        abs_eta,
-        decay_mode,
-        gen_match,
-        variation_efake_dm0_barrel,
-        variation_efake_dm1_barrel,
-        variation_efake_dm0_endcap,
-        variation_efake_dm1_endcap,
-        variation_mufake,
-        variation_gentau_dm0,
-        variation_gentau_dm1,
-        variation_gentau_dm10,
-        variation_gentau_dm11
-    );
-
     auto correction_lambda =
         [
             evaluator,
             id_algorithm,
             id_vs_jet_wp,
             id_vs_ele_wp,
-            variation
+            variation_efake_dm0_barrel,
+            variation_efake_dm1_barrel,
+            variation_efake_dm0_endcap,
+            variation_efake_dm1_endcap,
+            variation_mufake,
+            variation_gentau_dm0,
+            variation_gentau_dm1,
+            variation_gentau_dm10,
+            variation_gentau_dm11
         ] (
             const ROOT::RVec<float> &pts,
             const ROOT::RVec<float> &etas,
@@ -266,6 +258,22 @@ PtCorrectionMC(
                 auto abs_eta = etas.at(i);
                 auto decay_mode = decay_modes.at(i);
                 auto gen_match = gen_matches.at(i);
+
+                // set the variation depending on the gen match, decay mode, and barrel/endcap region
+                std::string variation = get_tes_variation(
+                    abs_eta,
+                    decay_mode,
+                    gen_match,
+                    variation_efake_dm0_barrel,
+                    variation_efake_dm1_barrel,
+                    variation_efake_dm0_endcap,
+                    variation_efake_dm1_endcap,
+                    variation_mufake,
+                    variation_gentau_dm0,
+                    variation_gentau_dm1,
+                    variation_gentau_dm10,
+                    variation_gentau_dm11
+                );
 
                 // evaluate the correction factor
                 // ensure that the tau fulfills the selection criteria for application of the correction,
--- a/src/taus.cxx
+++ b/src/taus.cxx
@@ -1287,6 +1287,9 @@ Id_vsMu(ROOT::RDF::RNode df,
  * scale factor and "up"/"down" for the up/down variation
  *
  * @return a new dataframe containing the new column
+ *
+ * @warning This function is deprecated. Use the overloaded function with the additional
+ * parameter `trigger_flag` instead.
  */
 ROOT::RDF::RNode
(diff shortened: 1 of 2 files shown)


src/taus.cxx
```diff
@@ -7,18 +7,346 @@
...
+ * @param variation_efake_dm0_endcap variation for electron faking a tau with
+ * decay mode 0 in the endcap region
+ * @param variation_efake_dm1_endcap variation for electron faking a tau with
+ * decay mode 1 in the endcap region
+ * @param variation_mufake variation for muon faking a tau
+ * @param variation_gentau_dm0 variation for genuine tau with decay mode 0
+ * @param variation_gentau_dm1 variation for genuine tau with decay mode 1
+ * @param variation_gentau_dm10 variation for genuine tau with decay mode 10
+ * @param variation_gentau_dm11 variation for genuine tau with decay mode 11
+ * @return the variation to use in the correctionlib evaluator
```
nshadskiy, line 38: Add an empty comment line between param and return line for style consistency.

src/taus.cxx
```diff
@@ -7,18 +7,346 @@
...
+    // debug information
+    Logger::get("physicsobject::tau::get_tes_variation")
+        ->debug(
+            "variation for tau abs eta {}, decaymode {}, gen match {}: "
+            "variation {}",
+            abs_eta,
+            decay_mode,
+            gen_match,
+            variation
+
```
nshadskiy, line 107: remove empty line

src/taus.cxx
```diff
@@ -7,18 +7,346 @@
...
+    const std::string &outputname,
+    const std::string &pt,
+    const std::string &eta,
+    const std::string &decay_mode,
+    const std::string &gen_match,
+    const std::string &es_file,
+    const std::string &correction_name,
+    const std::string &id_algorithm,
+    const std::string &id_vs_jet_wp,
+    const std::string &id_vs_ele_wp,
```
nshadskiy, line 207: I suggest to move these two parameters to the end of the inputs and directly setting the default values. By that run2 analyses don't need to specify these parameters with `""` by hand.

src/taus.cxx
```diff
@@ -7,18 +7,346 @@
...
+                // ensure that the tau fulfills the selection criteria for application of the correction,
+                // set the correction factor to 1 otherwise
+                float correction_factor = 1.0;
+                if (
+                    (decay_mode == 0)
+                    || (decay_mode == 1)
+                    || (decay_mode == 2)
+                    || (decay_mode == 10)
+                    || (decay_mode == 11)
+                ) {
```
nshadskiy, line 286: Can you try 
```cpp
const std::unordered_set<int> valid_modes = {0, 1, 2, 10, 11};
if (valid_modes.count(decay_mode)) {...}
```
to make this shorter.

src/taus.cxx
```diff
@@ -65,6 +393,10 @@ namespace tau {
  * originating from prompt electrons (`gen_match=1`) and electrons that decayed
  * from a tau lepton
  * (`gem_match=3`).
+ * 
+ * @note This function is intended to be used for Run 2 analyses. In Run 3,
+ * the tau energy scale corrections also depend on the DeepTau working points
+ * for ID vs. electrons and vs. jets. Use `TauCorrectionMC` instead.
```
nshadskiy, line 399: change to `physicsobject::tau::PtCorrectionMC`

src/taus.cxx
```diff
@@ -733,6 +1082,102 @@ Id_vsEle(ROOT::RDF::RNode df,
...
+ * ------------------------------------|-------|-------
+ * no ID selection (takes every tau)   |  0    | -
+ * VVVLoose                            |  1    | 1
+ * VVLoose                             |  2    | 2
+ * VLoose                              |  4    | 3
+ * Loose                               |  8    | 4
+ * Medium                              |  16   | 5
+ * Tight                               |  32   | 6
+ * VTight                              |  64   | 7
+ * VVTight                             |  128  | 8
```
nshadskiy, line 1104: Its not a bitmask anymore, right? These are integer values now (above nanoAODv12), so I would remove the part with the bitmasks.

src/taus.cxx
```diff
@@ -191,6 +523,10 @@ PtCorrectionMC_eleFake(ROOT::RDF::RNode df,
  * originating from prompt muons (`gen_match=2`) and muons that decayed from a
  * tau lepton
  * (`gem_match=4`).
+ *  *
```
nshadskiy, line 526: typo `*   *`

nshadskiy (approved):
Looks good to me! Can be merged after addressing the comments.

src/taus.cxx
moritzmolch: This is primarily to avoid errors when reading the correctionlib values. I included all, which are allowed in general, including decay mode 2.

src/taus.cxx
moritzmolch: I just keep the last column now.

src/taus.cxx
```diff
@@ -7,18 +7,346 @@
...
+                    variation_efake_dm0_barrel,
+                    variation_efake_dm1_barrel,
+                    variation_efake_dm0_endcap,
+                    variation_efake_dm1_endcap,
+                    variation_mufake,
+                    variation_gentau_dm0,
+                    variation_gentau_dm1,
+                    variation_gentau_dm10,
+                    variation_gentau_dm11
+                );
```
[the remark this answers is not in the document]
nshadskiy: I see, that's fine then.

PR #329, opened 2025-07-21
 Functions for jet corrections in Run 3

moritzmolch:
The procedure for the jet energy resolution corrections has slightly changed. The scale factor now also depends on the p<sub>T</sub> of the jet. This change has been accounted for by adding a parameter `lhc_run` to the `PtCorrectionMC` function, which has a default value of `2`. For a value of `3`, an adapted call for obtaining the jet energy resolution scale factor is used. For a value of `2`, the default behavior for Run 2 is be invoked.

### DIFF
--- a/src/jets.cxx
+++ b/src/jets.cxx
@@ -85,7 +85,7 @@ PtCorrectionMC(ROOT::RDF::RNode df,
                const std::vector<std::string> &jes_shift_sources,
                const std::string &jer_tag, bool reapply_jes,
                const int &jes_shift, const std::string &jer_shift,
-               const int jer_seed) {
+               const int jer_seed, const int& lhc_run = 2) {
     // In nanoAODv12 the type of jet/fatjet ID was changed to UChar_t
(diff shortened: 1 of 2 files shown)


src/jets.cxx
```diff
@@ -85,7 +85,7 @@ PtCorrectionMC(ROOT::RDF::RNode df,
                const std::vector<std::string> &jes_shift_sources,
                const std::string &jer_tag, bool reapply_jes,
                const int &jes_shift, const std::string &jer_shift,
-               const int jer_seed) {
+               const int jer_seed, const int& lhc_run = 2) {
```
nshadskiy, line 88: To stay consisted, can you add here `jer_seed = 42` and add the `lhc_run` parameter to the documentation.

nshadskiy (approved):
Looks good! Only one comment.

src/jets.cxx
moritzmolch: Done!

PR #330, opened 2025-07-22
Additional muon trigger scale factor function

moritzmolch:
This pull request adds a function for trigger scale factor evaluation that takes the trigger flag as additional input column. The scale factor is only evaluated when the trigger flag is set to true, i.e., when the trigger has fired. Otherwise, the scale factor is set to 1. This mitigates implementing acceptance conditions of the trigger directly in the C++ function.

### DIFF
--- a/src/muons.cxx
+++ b/src/muons.cxx
@@ -383,7 +383,7 @@ Trigger(ROOT::RDF::RNode df,
 
 /**
  * @brief This function calculates muon trigger scale factors (SFs) for a single
- * muon dependening on its pseudorapidity (\f$\eta\f$) and transverse momentum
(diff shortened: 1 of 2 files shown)


src/muons.cxx
```diff
@@ -375,6 +375,78 @@ Trigger(ROOT::RDF::RNode df,
         {pt, eta});
     return df1;
 }
+
+/**
+ * @brief This function calculates muon trigger scale factors (SFs) for a single
+ * muon dependening on its pseudorapidity (\f$\eta\f$) and transverse momentum
```
nshadskiy, line 381: *depending

PR #332, opened 2025-08-05
Jet ID column and TrigObj_filterBits type casting for NAnoAOD v15 compatibility

sofiagiappichini:
This PR introduces a new producer `physicsobject::jet::JetID` to calculate the jet ID for NanoAOD v15 samples from the dedicated `correctionlib` json file. It uses these evaluations to assign a jet ID to each jet in the input dataframe:
- 6 : passes Tight and Tight Lepton Veto IDs
- 2 : passes Tight ID but fails Tight Lepton Veto ID
- 0 : fails Tight ID
Ref. https://twiki.cern.ch/twiki/bin/view/CMS/JetID13p6TeV#Recommendations_for_the_13_6_AN1

Type casting for TrigObj_filterBits is also introduced in the corresponding producers since the type changed from previous versions (int) to v15 (ULong64_t).


src/pairselection.cxx
```diff
@@ -1230,8 +1230,8 @@ namespace mutau {
     - muon_phi
     - muon_mass
     - muon_iso
-    - tau_mask containing the flags whether the tau is a good tau or not
     - muon_mask containing the flags whether the muon is a good muon or not
+    - tau_mask containing the flags whether the tau is a good tau or not
```
nshadskiy, line 1234: Was this change intended?

sofiagiappichini: I didn't realise it made it to the PR, but I adjusted it to the order the parameters need to be passed to the producer

include/jets.hxx
```diff
@@ -4,6 +4,21 @@
...
+              const std::string &jet_eta,
+              const std::string &jet_chHEF,
+              const std::string &jet_neHEF,
+              const std::string &jet_chEmEF,
+              const std::string &jet_neEmEF,
+              const std::string &jet_muEF,
+              const std::string &jet_chMult,
+              const std::string &jet_neMult,
+              const std::string &jet_id_file,
+              const std::string &jet_name);
```
nshadskiy, line 21: I suggest to change the namespace of this function. Since this is a jet quantity, I would move it to: `physicsobject::jet::quantity::ID` with the function name `ID`. Similar to the tau IDFlag quantity.

build/.gitkeep
```diff
@@ -1 +0,0 @@
-build
```
nshadskiy, line 1: Did this cause you any issues?

src/triggers.cxx
```diff
@@ -233,17 +233,19 @@ ROOT::RDF::RNode GenerateSingleTriggerFlag(
...
                             const ROOT::Math::PtEtaPhiMVector &particle_p4,
-                            ROOT::RVec<int> triggerobject_bits,
+                            ROOT::RVec<ULong64_t> triggerobject_bits,
                             ROOT::RVec<UShort_t> triggerobject_ids_v12,
                             ROOT::RVec<float> triggerobject_pts,
                             ROOT::RVec<float> triggerobject_etas,
                             ROOT::RVec<float> triggerobject_phis) {
         auto triggerobject_ids = static_cast<ROOT::RVec<int>>(triggerobject_ids_v12);
+        // In nanoAODv15 the type of trigger bits was changed to ULong64_t
+        auto triggerobject_bits_int = static_cast<ROOT::RVec<int>>(triggerobject_bits);
```
nshadskiy, line 248: You also need to add this `utility::Cast <ROOT::RVec<ULong64_t>, ROOT::RVec<Int_t>>(...)` line for the `triggerobject_bit`, otherwise this will not work for nanoAOD v9 and v12.

src/jets.cxx
```diff
@@ -765,6 +765,106 @@ Btagging(ROOT::RDF::RNode df,
...
+namespace quantity {
+
+/**
+ * @brief Applies jet identification criteria based on JSON-defined jet ID corrections.
+ *
+ * This function loads jet ID definitions from correctionlib JSON files for the specified
+ * jet collection and evaluates two sets of criteria:
+ *  - Tight ID
+ *  - Tight Lepton Veto ID
+ *
```
nshadskiy, line 778: Can you move the quantity namespace before the scalefactor namespace, so that it matches the order in the .hxx file.

PR #334, opened 2025-08-07
Update to LCG 108 stack with cpp 15

sofiagiappichini:
This PR updates the environment to the newest LCG stack and cpp available. Some changes in the ONNX and ROOT paths are introduced, alongside explicitly adding Minuit as ROOT package. All other dependencies are unchanged.


a-monsch (approved):

moritzmolch (approved):
Running the code worked for me after merging this branch with the current `main`.

PR #337, opened 2025-08-20
Length of a vector quantity

moritzmolch:
This pull request adds a function that calculates the length of `ROOT::VecOps::RVec<T>` objects in a column and saves it to a new one. The function is included as template `physicsobject::Size<T>`, where `T` is the underlying data type of the `RVec` , i.e., for a column containing `RVec<int>` objects, `physicsobject::Size<int>` must be called.

This function should replace occasions of `physicsobject::Count` when the collection is represented by an index list instead of a boolean mask.

### DIFF
--- a/include/physicsobjects.hxx
+++ b/include/physicsobjects.hxx
@@ -302,6 +302,28 @@ CutQuantity(ROOT::RDF::RNode df, const std::string &outputname,
         {quantity});
 }
 
+/**
+ * @brief This function returns the length of the vector containers in columns
+ * that contain `ROOT::VecOps::RVec<T>` objects. The function is templated
+ * with `T`, which must be set to the underlying type of the vector object.
+ *
+ * @param df input dataframe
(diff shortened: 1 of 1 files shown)


include/physicsobjects.hxx
```diff
@@ -302,6 +302,28 @@ CutQuantity(ROOT::RDF::RNode df, const std::string &outputname,
...
+ *
+ * @param df input dataframe
+ * @param outputname name of the output column storing the object count
+ * @param vector_quantity name of the quantity stored in vector structures
+ *
+ * @return a dataframe with a new column
+ */
+template <typename T>
+ROOT::RDF::RNode Size(ROOT::RDF::RNode df, const std::string &outputname,
+                       const std::string &vector_quantity) {
```
nshadskiy, line 318: If you define it in a header and the function is small, it is better to define it with `inline`.
```cpp
inline ROOT::RDF::RNode ...
```

PR #341, opened 2025-08-28
Extend range of eta values for tau ID vs muon scale factors

moritzmolch:
For the last bin of the tau ID vs muon scale factors, the upper bin edge for eta has been increased from 2.3 to 2.4 to fully cover the acceptance range of taus in Run 3 (|eta| < 2.5). Run 2 analyses should not be affected by this, as the input bins already have a cut on |eta| < 2.3.


moritzmolch:
@a-monsch I think that this change is not affecting the Run 2 analysis, because the eta range of the input objects is already restricted to 2.3. But maybe I missed something.

a-monsch (approved):

PR #367, opened 2026-03-13
Update kingmaker.rst with connection issue note

jhvoss:
Added docs about potential connection issues with the remote scheduler and the use of set_luigiport function (PR https://github.com/KIT-CMS/KingMaker/pull/77)

### DIFF
--- a/docs/sphinx_source/kingmaker.rst
+++ b/docs/sphinx_source/kingmaker.rst
@@ -35,6 +35,9 @@ The setup script has also additional options:
     -l, --list                List available workflows
     -h, --help                Show this help message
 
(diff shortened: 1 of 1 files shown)


tvoigtlaender (approved):

docs/sphinx_source/kingmaker.rst
```diff
@@ -35,6 +35,9 @@ The setup script has also additional options:
     -l, --list                List available workflows
     -h, --help                Show this help message
 
+It is possible that the session failes to connect to the remote scheduler after some time. The ```set_luigiport``` function can be used to connect the scheduler to a new port (see PR https://github.com/KIT-CMS/KingMaker/pull/77).
+
+
```
nshadskiy, line 40: I think this would have needed an execution example, on first read I didn't get that I can just run this new function in the terminal. 
Also not sure if it makes that much sense to give the plan link to  the PR because there this functionality is also not described.

PR #372, opened 2026-04-20
CROWN Changes for mumu to ee embedding 

FFA76:


code_generation/configuration.py
```diff
@@ -159,6 +159,18 @@ def _set_sample_parameters(self) -> None:
...
+        else:
+            sample_parameters["is_data"] = False
+        if "dyjets" in self.sample:
+            sample_parameters["is_dyjets"] = True
+        else: 
+            sample_parameters["is_dyjets"] = False
+        if "wjets" in self.sample:
+            sample_parameters["is_wjets"] = True
+        else:
+            sample_parameters["is_wjets"] = False
```
nshadskiy, line 173: Why is this needed? This is already done by the if else above.

code_generation/configuration.py
```diff
@@ -636,7 +648,7 @@ def _remove_empty_configkeys(self, config) -> None:
             Returns:
                 None
         """
-        for key in config:
+        for key in list(config.keys()):
```
nshadskiy, line 651: Was this breaking anything?

src/event.cxx
```diff
@@ -68,10 +70,117 @@ GenerateSeed(
...
+    ROOT::RDF::RNode df,
+    const std::string &outputname,
+    const std::string &reco_q,
+    const std::string &emb_q_1,
+    const std::string &emb_q_2) {
+    auto has_match = [](const int &reco_q,
+                        const float &emb_q_1,
+                        const float &emb_q_2) {
+        return static_cast<int>(reco_q == static_cast<int>(emb_q_1) ||
+                                reco_q == static_cast<int>(emb_q_2));
```
nshadskiy, line 149: Is this ever false?

src/event.cxx
```diff
@@ -68,10 +70,117 @@ GenerateSeed(
...
+ * @param q_2 name of the second reconstructed charge column
+ *
+ * @return a dataframe with the new boolean flag column
+ */
+ROOT::RDF::RNode HasOppositeCharge(
+    ROOT::RDF::RNode df,
+    const std::string &outputname,
+    const std::string &q_1,
+    const std::string &q_2) {
+    auto has_opposite_charge = [](const int &q_1, const int &q_2) {
```
nshadskiy, line 174: Why are here the charges integer but in the other functions floats?

code_generation/configuration.py
moritzmolch: Agree. I would keep the old version.

include/event.hxx
```diff
@@ -327,6 +327,31 @@ GenerateSeed(
     const UInt_t &master_seed = 42
 );
 
+// Select the embedding-muon p4 that has the same charge as the reco object.
```
moritzmolch, line 330: Please run clang-format (https://github.com/KIT-CMS/CROWN/blob/main/checks/cpp-formatting.sh). Merging the branch will be blocked because the code formatting test is not going to pass.

src/event.cxx
```diff
@@ -68,10 +70,117 @@ GenerateSeed(
...
+ * @param outputname name of the new column containing the matched embedding muon Lorentz vector
+ * @param reco_q name of the reconstructed particle charge column
+ * @param emb_q_1 name of the leading embedding muon charge column
+ * @param emb_q_2 name of the trailing embedding muon charge column
+ * @param emb_p4_1 name of the leading embedding muon Lorentz vector column
+ * @param emb_p4_2 name of the trailing embedding muon Lorentz vector column
+ *
+ * @return a dataframe with the new Lorentz vector column
+ */
+ROOT::RDF::RNode MatchEmbeddingMuonP4ByCharge(
```
moritzmolch, line 94: The name of the function suggests that a matching takes place and that a boolean value or the matching metric is returned. A name like `GetEmbeddingMuonP4ByCharge` suits better here.

src/event.cxx
moritzmolch: Agree. The embedding selection already imposes opposite-sign charge muons, so these conditions should always be fulfilled. However, the function could be useful for consistency checks.

moritzmolch (changes requested):
Only a few minor comments.

Consider rebasing main into this branch (to keep up-to-date with the changes in main since you have branched off).

PR #369, opened 2026-03-31
Add DAG visualization to CROWN configs

tvoigtlaender:
Adds an html-based DAG inspection tool and a python function to create the json data necessary for the tool.
Includes search functionality and input/output information on click of nodes/edges.

Intermediate goals:

- [x] Separation by scope
- [x] Replacement of I/O nodes
- [x] drop down menus for scope, era, and sample type
- [x] improved scope ordering (global above non-global)
- [x] Type coding of nodes by purpose (cut, define, ...)
- [x] Jump to option for Input/output elements at the edges/nodes
- [x] Extended producer information on click
- [x] Compatibility with friends

PR #378, opened 2026-05-07
Fix data structure of metadata in generated CROWN Ntuples

tvoigtlaender:
Use single object to save full metadata dict over group of branches.
Add metadata to quantities maps.
Require quantitiesmap files to match analysis/main-config/era/sample_type with compiling friend-config.
Expose origin config of quantities during friend compilation for use with DAG visualization (https://github.com/KIT-CMS/CROWN/pull/369)


code_generation/friend_trees.py
```diff
@@ -240,26 +243,33 @@ def _readout_input_json_file(
...
                 f"Era {self.era} not found in input information file {input_file}.\n"
             )
-            errorstring += f"Available eras are: {data.keys()}"
+            errorstring += f"Available eras are: {quantity_data.keys()}"
             raise ConfigurationError(errorstring)
-        if self.sample not in data[self.era].keys():
+        if (
+            self.sample not in quantity_data[self.era].keys()
+            or self.sample != metadata["sample_type"]
+        ):
```
nshadskiy, line 258: Check if this should better be an `and` instead of `or`.

tvoigtlaender: Generally both of the conditions should be satisfied here, but the error message is misleading.

When 'self.era != metadata["era"]', then there is a mismatch between the creation era of the ntuple file and the usage era for the new friend.
A split into 2 errors would make sense here.

same for the sample type.

nshadskiy: I misunderstood the logic a bit, `or` is fine since I expect only one of them to be wrong and then trigger the error message.

nshadskiy (approved):

PR #381, opened 2026-05-20
Handling of multiple trigger bits to be matched

sofiagiappichini:
Trigger bit values need to be matched as a full list in double triggers. This required a reworking of the current producers, and in particular the matching of the trigger with the offline object (`matchParticle`).  Because of that, both single and double trigger producers (`SingleObjectFlag` and `DoubleObjectFlag`) are updated to take a vector of int as trigger bit values. If the vector has only one entry, the previous behaviour is maintained. Otherwise, all values in the vector need to be matched simultaneously by the object, as instructed in the [Tau Twiki](https://twiki.cern.ch/twiki/bin/view/CMS/TauTrigger#Tau_Triggers_in_NanoAOD_general).


src/jets.cxx
```diff
@@ -724,8 +724,7 @@ PtCorrectionMC(ROOT::RDF::RNode df,
             float reso =
                 jet_energy_resolution(etas.at(i), corrected_pts.at(i), rho);
             float reso_sf = 1.0;
-            if (std::stoi(era.substr(0, 4)) <= 2018) {
-                // run 2 case
+            if (std::stoi(era.substr(0, 4)) <= 2018) { // run 2 case
                 reso_sf = jer_sf_evaluator->evaluate({etas.at(i), jer_shift});
             } else {
                 // run 3 case
```
moritzmolch, line 731: For consistent style, move comment `// run 3 case` to the line before (similar to the change in lines 727-728).

moritzmolch (commented):
Looks fine, I only have one style comment.

code_generation/configuration.py
```diff
@@ -159,6 +159,18 @@ def _set_sample_parameters(self) -> None:
...
+        else:
+            sample_parameters["is_data"] = False
+        if "dyjets" in self.sample:
+            sample_parameters["is_dyjets"] = True
+        else:
+            sample_parameters["is_dyjets"] = False
+        if "wjets" in self.sample:
+            sample_parameters["is_wjets"] = True
+        else:
+            sample_parameters["is_wjets"] = False
```
nshadskiy: Why should this be needed? There is the loop above doing the same thing.

sofiagiappichini: Seems like we don't, this was a previous version of the implementation that I had on my branch that didn't get replaced by the subsequent merges. You can remove it.

nshadskiy (approved):

PR #384, opened 2026-06-01
Fixes to Tau, Jet, Top Pt producers

sofiagiappichini:
Changes in this PR:

- A new Top pt reweighting function is introduced for Run3, following the recommendation by TOP PAG
- Jet ES uncertainties are simplified into a single string instead of a vector, since they are always passed individually
- Tau pt correction, ID vs Jet, Ele, and Mu producers are created for Run3/NanoAOD >=v12 with a simplified uncertainty scheme, as suggested by TAU POG. Plus, in all producers, eta is now passed as the absolute value as intended, and in the new Tau pt producer the selected decay modes are passed as arguments instead of being hardcoded.


init.sh
tvoigtlaender, line 1: Doesn't this break CROWN usage with KingMaker?
It kind of relies on a conda env.
At least check if the current conda env is the base conda env.
Also, this issue may be user-specific.

sofiagiappichini: Actually, I just ran some tests, and if you source KingMaker before CROWN, then CROWN breaks even more, i.e. it cannot find the samples when they clearly exist, or there are issues of library version mismatch. It only works when no environment is present before. If other users don't ever see this issue (after `source init.sh -c lcg`), then I will remove the change.

src/jets.cxx
jhvoss: here it's still the 'old' style and I can't find a different overloaded version with the new style and same inputs

init.sh
tvoigtlaender: Unfortunately, I'm not sure how the lcg env is currently implemented as @nshadskiy  had worked on it last.

nshadskiy: The only possible issue I see is when KingMaker (with CROWN) is newly set up. Then the KingMaker env is used as well as the `init.sh` script. This needs to be checked.

nshadskiy: After some tests, I would say this change is save. During a fresh setup, a subshell is used so the initial KingMaker environment is not affected.

nshadskiy (approved):

PR #385, opened 2026-06-11
Switch std::cerr and std::cout to Logger for FastMTT code

nshadskiy:
The `std::cerr` can causing problems with the buffer when running in a job, therefore, the prints are now moved to the Logger and can be printed it running in debug mode.

Additionally, a small bin loop issues is fixed.


jhvoss:
Looks good :+1:

jhvoss (approved):

PR #388, opened 2026-07-15
Improve producer setup

nshadskiy:
# PR: Improve Producer Setup with Producer switching and Context-Aware Defaults

## Summary
Introduces producer switching for era/nanoAOD-dependent variants and a context manager system for simplifying producer configuration by reducing boilerplate code.

Related to https://github.com/KIT-CMS/TauAnalysis-CROWN/blob/main/scripts/CROWNWrapper.py and https://github.com/KIT-CMS/TauAnalysis-CROWN/issues/82

---

### 1. Producer Switching (producer.py)
- **`SwitchProducer`** base class for managing era/version-dependent producer variants
- **`RUN2_ERAS`**: Classifies 2016-2018 as Run 2, all others as Run 3
- **`ERA_NANOAOD_VERSION_DEFAULTS`**: Maps eras to default nanoAOD versions (v9/v12/v15)

**Usage:**
```python
class JetEnergyCorrection(SwitchProducer):
    class run2:
        v9 = ProducerGroup(subproducers=[...])
    class run3:
        v12 = ProducerGroup(subproducers=[...])
        v15 = ProducerGroup(subproducers=[...])

# Auto-select based on era
jets.JetEnergyCorrection.get(era)
```

### 2. Context-Aware Defaults (helpers.py)
- **`defaults()` context manager** for setting default parameter values
- All producer constructors now accept optional parameters with auto-detection
- Automatic name detection from variable assignment

**Usage:**
```python
with defaults(scopes=['em']):
    ElectronProducer = Producer(
        call='correctElectron({input})',
        input=[electron_pt],
        output=[electron_pt_corrected]
    )
    
    MuonProducer = Producer(
        call='correctMuon({input})',
        input=[muon_pt],
        output=[muon_pt_corrected]
    )
```

### 3. Enhanced Systematics (`systematics.py`)
- **`get_add_shift()`** factory function for simplified shift configuration
- Supports both simple and complex shift configurations via context manager


code_generation/producer.py
```diff
@@ -466,18 +524,20 @@ def writecalls(
...
-        output: str,
-        scope: Union[List[str], str],
-        vec_config: str,
+        name: Union[str, None] = None,
+        call: Union[str, None] = None,
+        input: Union[  # pylint: disable=redefined-builtin
+            List[q.Quantity], Dict[str, List[q.Quantity]], None
+        ] = None,
+        output: Union[str, None] = None,  # pylint: disable=redefined-builtin
+        scope: Union[List[str], str, None] = None,
```
a-monsch: Is the differentiation to "scope" instead of "scopes" that is used everywhere else intentional?

If for backcompatibility reasons it stays `scope`, then an introduction of "scopes" (and a check that only one of both is provided) would be reasonable, since:

1. `CONTEXT_REGISTRY` asks for `"scopes"`
2. If users would move such a Producer in/out of a default context, they would either move `scopes` into `__init__` which would not work (or move `scope` in the context manager, which would do nothing as well).

See herefore my comment regarding this inconsistency [here](https://github.com/KIT-CMS/TauAnalysis-CROWN/blob/5afa97374b506aa6bb6e70df4562cae692df1f38/scripts/CROWNWrapper.py#L246) :D

code_generation/quantity.py
```diff
@@ -2,13 +2,14 @@
...
+from code_generation.helpers import get_variable_name
 
 log = logging.getLogger(__name__)
 
 
 class Quantity:
-    def __init__(self, name: str):
-        self.name = name
+    def __init__(self, name: Union[str, None] = None):
+        self.name = name or get_variable_name()
```
a-monsch, line 29: A docstring regarding this auto-detection feature would be needed here, similarly to the producers as well?

code_generation/producer.py
```diff
@@ -23,51 +29,72 @@ def __missing__(self, key: Any) -> Any:
...
             output: A list of output quantities
             scopes: A list of scopes in which the producer is used
             is_filter: True if the producer is an event filter
 
         """
+        # Auto-detect name if not provided
+        name = name or get_variable_name()
+
+        # Get context-aware defaults
+        _call = call if call is not None else CONTEXT_REGISTRY["call"].get()
```
a-monsch, line 59: given that above

`name = name or get_variable_name()`

Is used, why switching here and in all subsequent places to an explicit `None` check? And not staying with `call or CONTEXT_REGISTRY["call"].get()`? As long as we are using python built-in container, this check should always work.

code_generation/helpers.py
```diff
@@ -22,3 +73,43 @@ def is_empty(value: Any) -> bool:
...
+        RuntimeError: If variable name cannot be determined from context
+    """
+    frame = inspect.currentframe().f_back.f_back
+    code_context = inspect.getframeinfo(frame).code_context
+    if code_context:
+        call_line = code_context[0].strip()
+        match = re.match(r"([\w\d_]+)\s*(=|:=)", call_line)
+        if match:
+            return match.group(1)
+    raise RuntimeError("Could not determine variable name from context")
```
a-monsch: This could be `NameNotDetermined`. Given that all subsequent name None checks could be omitted?

code_generation/producer.py
```diff
@@ -466,18 +524,20 @@ def writecalls(
...
-        scope: Union[List[str], str],
-        vec_config: str,
+        name: Union[str, None] = None,
+        call: Union[str, None] = None,
+        input: Union[  # pylint: disable=redefined-builtin
+            List[q.Quantity], Dict[str, List[q.Quantity]], None
+        ] = None,
+        output: Union[str, None] = None,  # pylint: disable=redefined-builtin
+        scope: Union[List[str], str, None] = None,
+        vec_config: Union[str, None] = None,
```
a-monsch, line 534: same story here

code_generation/producer.py
```diff
@@ -984,3 +1146,112 @@ def CollectProducerOutput(
...
+            ValueError: When no matching variant is found.
+        """
+        run = "run2" if era in RUN2_ERAS else "run3"
+        obj = getattr(cls, run) if hasattr(cls, run) else cls
+
+        # If the run attribute is already a producer, return it directly
+        if not isinstance(obj, type):
+            return obj
+
+        target_version = (
```
a-monsch: Similar discussion to `name or _get_variable_name()` from before here and in the if statement below

code_generation/producer.py
```diff
@@ -984,3 +1146,112 @@ def CollectProducerOutput(
...
+    2. If the class has a ``run2`` / ``run3`` attribute, descend into it;
+       otherwise operate on the class itself.
+    3. If the resulting object is already a producer (not a nested class),
+       return it directly.
+    4. Look up *version* (or the era default from :data:`ERA_NANOAOD_VERSION_DEFAULTS`)
+       as an attribute of the object.
+    """
+
+    @classmethod
+    def get(cls, era: str, version: Union[str, None] = None) -> Any:
```
a-monsch: the output type here would be a `Union[Producer, Filter, <all other possibilities>]` and not `Any`.

code_generation/quantity.py
```diff
@@ -251,8 +252,8 @@ class NanoAODQuantity(Quantity):
     are therefore shielded from using them directly as a output.
     """
 
-    def __init__(self, name: str):
-        super().__init__(name)
+    def __init__(self, name: Union[str, None] = None):
+        super().__init__(name or get_variable_name())
```
a-monsch, line 295: Similar comment as in `Quantity`

code_generation/systematics.py
```diff
@@ -11,10 +11,124 @@
...
+        _shift_map = shift_map or CONTEXT_REGISTRY["shift_map"].get()
+        _producers = producers or CONTEXT_REGISTRY["producers"].get()
+
+        to_be_checked = (
+            [_name, shift_config, _producers]
+            if shift_config is not None and isinstance(_producers, dict)
+            else [_name, _scopes, _shift_key, _shift_map, _producers]
+        )
+
+        if any(it is None for it in to_be_checked):
```
a-monsch: This can be translated to a for loop like check as was done for the producers to be more transparent for the users about what is missing?

a-monsch (commented):
For completeness:

  * This builds upon the existing feature in [TauAnalysisCROWN](https://github.com/KIT-CMS/TauAnalysis-CROWN/blob/main/scripts/CROWNWrapper.py) and a [feature request](https://github.com/KIT-CMS/TauAnalysis-CROWN/issues/82) therein. Hence, when this PR merges the implementation of this feature needs to be removed there in order to work and the feature-request closed accordingly.
  * Since this is a major feature addition this should be propagated to the official documentation?
  * This can also be explained in `templace_analysis` that is used for the CROWN onboarding?

code_generation/producer.py
nshadskiy: I did this in the beginning, however this lead to errors that some of the parameters being badly defined.

code_generation/helpers.py
nshadskiy: Thats how it was in the wrapper, but I can change that.

code_generation/producer.py
nshadskiy: I changed it to `scopes` in all producers. This needs to be adjusted by the users in their analyses but I think this is still better in terms of consistency that any other solution.

code_generation/quantity.py
nshadskiy: Done

code_generation/helpers.py
nshadskiy: Done

code_generation/producer.py
nshadskiy: Regarding the `vec_config` and `vec_configs`, there is a technical difference between the two producers they are used in (the first one cannot be a list) and they are appearing only in those producers (not like scopes), therefore, I would like to keep the names different.

code_generation/producer.py
nshadskiy: Done

code_generation/quantity.py
nshadskiy: Done

code_generation/systematics.py
nshadskiy: Done

code_generation/producer.py
nshadskiy: Done

analysis_configurations/template_analysis/producers/event.py
```diff
@@ -1,90 +1,73 @@
...
-)
-is_diboson = Producer(
-    name="is_diboson",
-    call="event::quantity::Define({df}, {output}, {is_diboson})",
-    input=[],
-    output=[q.is_diboson],
-    scopes=["global"],
-)
+    with defaults(input=[]):
+        is_data = Producer(
```
a-monsch, line 28: an off-topic question: why do we actually need those flags in the template analysis

analysis_configurations/template_analysis/producers/event.py
```diff
@@ -1,90 +1,73 @@
...
-    name="MetFilter",
-    call='event::filter::Flag({df}, "{met_filters}", "{met_filters}")',
-    input=[],
-    output=None,
-    scopes=["global"],
-    vec_configs=["met_filters"],
-)
+    MetFilter = VectorProducer(
+        call='event::filter::Flag({df}, "{met_filters}", "{met_filters}")',
+        input=[],
```
a-monsch: this can also go under the input=[] default?

a-monsch (commented):
Looks good. minute changes for some template_analysis producers proposed in https://github.com/KIT-CMS/CROWN/pull/391

analysis_configurations/template_analysis/producers/event.py
nshadskiy: They are not necessary, but just to have some producers that do something I would say.

a-monsch (approved):

PR #390, opened 2026-07-22
Muon pt corrections and other systematics 

sofiagiappichini:
This PR introduces Run 3 muon pt corrections following the provided instructions by the MUO POG for scale and resolution. A new producer is added to take care of MET unclustered energy corrections, following JME recommendations. The Golden JSON for 2026 data is added.

Fixes:
- in `tau.cxx` allows for 2026 era
- btag systematics for the fixed WP producers are now split by jet flavor as recommended by BTV
- muon ID and isolation producers are merged into one since the logic is identical
- electron pt corrections are now fully relying on correctionlib instead of deriving parameters in the code as recommended by EGM
- support of new JER JSONS with the introduction of `SFUncertianty` and minor fix to 2024 logic
- modification to recoil corrections so that an invalid variation (i.e. `nom`) returns the input MET - needed for upcoming PR to the TauAnalysis side


src/met.cxx
```diff
@@ -666,6 +660,74 @@ Type1Correction(ROOT::RDF::RNode df, const std::string &outputname,
...
+ * NanoAOD MET \f$p_T\f$
+ * @param met_phi_nominal name of the column with the nominal (never shifted)
+ * NanoAOD MET \f$\phi\f$
+ * @param met_pt_shifted name of the column with the NanoAOD MET \f$p_T\f$,
+ * which is equal to `met_pt_nominal` if no unclustered energy shift is
+ * active, or to the Unclustered Up/Down variant otherwise
+ * @param met_phi_shifted name of the column with the NanoAOD MET \f$\phi\f$,
+ * analogous to `met_pt_shifted`
+ * @param apply_shift if set to true, the shift is applied; if false, the
+ * output column is just the unmodified input MET vector
```
nshadskiy: This parameter is not used.

src/met.cxx
```diff
@@ -177,18 +178,11 @@ RecoilCorrection(ROOT::RDF::RNode df,
...
                                     variation);
-                        throw std::runtime_error(
-                            "Invalid variation for Recoil corrections");
+                        met_new = met;
                     }
-                } else if (method == "QuantileMapFit") {
-                    Logger::get("met::RecoilCorrection")
-                        ->debug("QuantileMapFit method not yet implemented, "
-                                "returning uncorrected MET");
-                    return met;
```
nshadskiy, line 191: Is this method not supported anymore by HLepRare?

src/muons.cxx
```diff
@@ -8,13 +8,243 @@
...
+ * @param shift name of the requested variation, "nom" for the nominal
+ * correction, "ScaleUp"/"ScaleDown" for the scale uncertainty, or
+ * "ResoUp"/"ResoDown" for the resolution uncertainty
+ * @param is_data flag that is `true` if the input is data and `false` if it
+ * is MC; resolution smearing is only applied to MC
+ *
+ * @return a dataframe with the new column
+ */
+ROOT::RDF::RNode
+PtCorrectionMC(ROOT::RDF::RNode df,
```
nshadskiy: Since it is for both MC and data, I would remove the MC in the function name.

src/met.cxx
sofiagiappichini: it's supported in the json but didn't write the code for it so that's why the message is there

src/muons.cxx
sofiagiappichini: agree, but also other producers for pt correction need adjusting i believe

src/muons.cxx
```diff
@@ -8,13 +8,243 @@
...
+
+            float a_f =
+                scale_evaluator_a->evaluate({eta.at(i), phi.at(i), "nom"});
+            float m_f =
+                scale_evaluator_m->evaluate({eta.at(i), phi.at(i), "nom"});
+
+            // apply scale correction to both data and mc
+            smear_pt = 1. / (m_f / pt.at(i) + charge.at(i) * a_f);
+
+            if (smear_pt < min_muon_pt || smear_pt > 200.0 ||
```
nshadskiy: Two things here, first, wouldn't it be more accurate to can it `scaled_pt` instead of `smear_pt` at this point? Second, I would also introduce a `max_muon_pt` if you have already done it for `min_muon_pt`.

src/muons.cxx
```diff
@@ -277,37 +457,52 @@ ROOT::RDF::RNode Id(ROOT::RDF::RNode df,
...
+                    sf = sf - evaluator->evaluate({std::abs(eta), pt, "syst"});
+                } else if (variation == "statup") {
+                    sf = sf + evaluator->evaluate({std::abs(eta), pt, "stat"});
+                } else if (variation == "statdown") {
+                    sf = sf - evaluator->evaluate({std::abs(eta), pt, "stat"});
+                } else {
+                    Logger::get("physicsobject::muon::scalefactor::IsoAndID")
+                        ->debug(
+                            "variation {} not implemented, check your code");
+                }
```
nshadskiy, line 565: For consistency this should also be added for the Reco and Trigger scale factors.

src/muons.cxx
sofiagiappichini: i don't agree with the second point, since going in pt>200 means the high pt region where the POG suggests a completely different way of applying corrections, while the minumum can be slightly adjusted to cover the analysis range

include/muons.hxx
```diff
@@ -28,16 +38,12 @@ ROOT::RDF::RNode Reco(ROOT::RDF::RNode df,
                       const std::string &outputname, const std::string &pt,
                       const std::string &eta, const std::string &sf_file,
                       const std::string &sf_name, const std::string &variation);
-ROOT::RDF::RNode Id(ROOT::RDF::RNode df,
```
moritzmolch, line 31: Why is it necessary to merge `Id` and `Iso` into a single function?

src/met.cxx
moritzmolch: I would switch `debug` to `error`, so that the message is also printed when you are not running in debug mode. Also, would it make sense to let the program crash at this point, so that analyzers are forced to revisit their analysis configuration?

include/muons.hxx
sofiagiappichini: Not necessary but since it's the same code, they can use the same producer.

src/met.cxx
sofiagiappichini: This case falls back to the next else and raises an error.

src/muons.cxx
sofiagiappichini: Add what? If you mean sf +- variation, this is only needed for iso and id, trigger and reco are already calculated fully.

include/muons.hxx
moritzmolch: Ok, now I understand, we still need to evaluate both corrections separately in the configuration. Fine. :+1:

src/muons.cxx
nshadskiy: This only applies for the medium muon pT case, but what if an analysis wants to use a the low or high pT correction. I would say the function should work for all cases.

src/muons.cxx
nshadskiy: Also for reco and trigger this should be updated to have the stat and syst split since this the general recommendation of the MuonPOG for their scale factors.

src/muons.cxx
sofiagiappichini: That is a completely different and extremely more complicated case to implement. If an analysis needs it, they can implement it.

nshadskiy (approved):
Thanks for doing all these updates @sofiagiappichini !

PR #392, opened 2026-07-29
Fix: Return empty quantity map for missing scopes or sample types in file

moritzmolch:
Return empty map if quantity map does not have scope or sample type instead of error and crash. This is needed if multiple scopes are processed and quantity maps only have a single scope.


nshadskiy (approved):
I agree with removing the errors. This will at least allow the users to be able to use the main branch again.
The general problem is a bit more complex including individual quantity map files for scopes and samples but also having it within the dicts in the files. A better approach would be to have a single quantity map file that includes all of the samples and scopes but this can be done in a separate PR.

PR #395, opened 2026-08-03
Rename of cut producers

sofiagiappichini:
This PR introduces some missing producers and restructures the names to be coherent everywhere. This means some boundaries in the analyses are now changed if the producers are not updated accordingly.

- `Max` means smaller or equal
- `Min` means greater or equal
- `Smaller` is strictly smaller
- `Greater` is strictly greater

Before `Max` meant smaller, while `Min` was greater equal, making them inconsistent. Test and template analyses are updated because of this to have the same logic.

### DIFF
--- a/analysis_configurations/template_analysis/producers/muons.py
+++ b/analysis_configurations/template_analysis/producers/muons.py
@@ -14,15 +14,15 @@ with defaults(scopes=["global"]):
             input=[nanoAOD.Muon_pt],
         )
         MuonEtaCut = Producer(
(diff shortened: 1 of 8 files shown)


moritzmolch (approved):
Looks fine.

nshadskiy (approved):
I agree, also covering all possible case is good even if they are not used often.
After the merge, analyzer will need to adjust their analysis setups if they use the main branch since these are breaking changes, but as far as I can see all were tagged for review so they should know.

PR #396, opened 2026-08-12
Improvements to producer modifications

sofiagiappichini:
Now the `ReplaceProducer`, `RemoveProducer`, and `AppendProducer` methods take `eras` and `exclude_eras` parameters in the same logic of `samples` and `exclude_samples`, both of which are also now optional, with a default that the method will be applied to all samples and eras if no list is ever specified.


nshadskiy (approved):
Looks good to me 👍

